PR #1, opened 2018-11-30
Update command names to remove 'advisory:' notation

smunilla:
Now that elliott is a standalone project, this notation makes less
sense. So let's all save the keystrokes and use the shorter notation.

### DIFF
--- a/elliott
+++ b/elliott
@@ -241,7 +241,7 @@ each function call.
 # Set advisory state
 # advisory:state
 #
-@cli.command("advisory:change-state", short_help="Change ADVISORY state")
+@cli.command("change-state", short_help="Change ADVISORY state")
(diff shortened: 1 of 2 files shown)


elliott
```diff
@@ -287,7 +287,7 @@ Bugzilla Bugs.
 # Create Advisory (RPM and image)
 # advisory:create
 #
-@cli.command("advisory:create", short_help="Create a new advisory")
+@cli.command("create-advisory", short_help="Create a new advisory")
```
tbielawa: lol

tbielawa:
Just curious about some consistency in renaming choices.

elliott
smunilla: I just missed that one. Fixing it.

PR #3, opened 2018-11-30
Remove unused runtime code

smunilla:
Removes all of the runtime code associated with cloneing distgits and
related to rebasing from OKD. None of this is needed from elliott functions
and is just a carryover from before the prokect split.


adammhaile:
YES. So much code removed :)

PR #7, opened 2018-12-04
Move errata tool template information out of code

smunilla:
This removes errata tool values from the source code and into an
ocp-build-data like datastore. In addition to moving everything to
a single source of truth, this will enable non-OpenShift teams to
use the errata tool functions of elliott in the future.


smunilla:
Depends on https://github.com/openshift/ocp-build-data/pull/25

PR #10, opened 2019-01-10
Transition to using python-bugzilla library

smunilla:
Move away from using shell callouts to the python-bugzilla cli and use the actual library. Also remove the deprecated bug transition searches now that we can add VERIFIED bugs regardless of the path they took to get there.

### DIFF
--- a/elliott
+++ b/elliott
@@ -21,7 +21,7 @@ import sys
 from elliottlib import version
 from elliottlib import Runtime
 import elliottlib.constants
-import elliottlib.bugzilla
+import elliottlib.bzutil
 import elliottlib.brew
 import elliottlib.errata
 import elliottlib.exceptions
@@ -31,6 +31,7 @@ from elliottlib.exceptions import ElliottFatalError
 from elliottlib.util import *
 
 # 3rd party
+import bugzilla
 import click
 import requests
 import dotconfig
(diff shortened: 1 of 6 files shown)


elliott
```diff
@@ -353,9 +354,10 @@ manually. Provide one or more --id's for manual bug addition.
         raise click.BadParameter("If not using --auto then one or more --id's must be provided")
 
     if auto:
-        bug_ids = elliottlib.bugzilla.search_for_bugs(bz_data, status)
+        bug_ids = elliottlib.bzutil.search_for_bugs(bz_data, status)
     else:
-        bug_ids = [elliottlib.bugzilla.Bug(id=i) for i in id]
+        bzapi = elliottlib.bzutil.get_bzapi(bz_data)
+        bug_ids = [bzapi.getbug(i) for i in id]
```
tbielawa: We are the best at duplicating code

elliottlib/bzutil.py
```diff
@@ -22,6 +23,7 @@ def search_for_bugs(bz_data, status, verbose=False):
 
     :return: A list of Bug objects
     """
+    bzapi = get_bzapi(bz_data)
```
tbielawa: When is this imported? Does that need the `bugzilla` module prefix?

tbielawa: My mistake. That's defined below in this file.

PR #12, opened 2019-01-16
Fix adding found builds to advisories

smunilla:


elliott
```diff
@@ -544,7 +543,7 @@ PRESENT advisory. Here are some examples:
         # Search and attach
         try:
             erratum = Erratum(errata_id=advisory)
-            erratum.addBuilds(unshipped_builds)
+            erratum.addBuilds([build.nvr for build in unshipped_builds], release = product_version)
```
tbielawa: idk why you have extra whitespace around the `=` sigh there. But ok.

tbielawa:
otherwise lgtm.

elliott
smunilla: Fixed.

PR #24, opened 2019-03-28
Fix further unsafe yaml loading.

ralphbean:
Any reason these weren't handled in #21?


tbielawa:
/lgtm

I imagine those were just overlooked

PR #30, opened 2019-04-03
Elliott can use a default advisory if defined in ocp-build-data

tbielawa:
To make bug sweeping happen more frequently and automatically we are
adding this little bit of functionality. If a value is defined in
ocp-build-data for .advisories.<TYPE> (type=rpm, image, security) then
you can have elliott attach bugs to that advisory by selecting it on
the command line.

We have played around with implementing logic to automatically
determine the right advisory for this but it hasn't come to fruition
yet. Instead we're going to try this little workaround instead.

----

Also adds some color to `find-bugs` and corrects a busted example.

### DIFF
--- a/elliottlib/schema_group.yml
+++ b/elliottlib/schema_group.yml
@@ -4,6 +4,19 @@
 type: map
 
 mapping:
+  # ID of the advisory we would auto attach bugs/builds to for the
+  # selected --group
+  "advisories":
+    type: map
+    mapping:
+      "image":
+        type: integer
+      "rpm":
+        type: integer
+      # security == 'RHSA'
+      "security":
+        type: integer
+
   "vars":
     type: map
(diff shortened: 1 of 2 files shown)


elliottlib/schema_group.yml
```diff
@@ -4,6 +4,19 @@
 type: map
 
 mapping:
+  # ID of the advisory we would auto attach bugs/builds to for the
+  # selected --group
+  "advisories":
```
sosiouxme: Actually, it occurs to me... Wouldn't this be better off in erratatool.yml? More topical and I don't think doozer needs it.

sosiouxme: Retracted; that's not what `erratatool.yml` is for. It's fine in `group.yml`.

PR #31, opened 2019-04-11
Fix repair-bugs command

smunilla:
Something changed with the BZ api and this wasn't working anymore. I also
added the ability to specify specific IDs if desired.


tbielawa:
/lgtm
seems legit

PR #32, opened 2019-04-16
Update find-bugs for payload diffs

smunilla:


elliott
```diff
@@ -335,10 +336,14 @@ advisory.
               metavar='ADVISORY_TYPE',
               type=click.Choice(['image', 'rpm', 'security']),
               help="Use the default value from ocp-build-data for ADVISORY_TYPE [image, rpm, security]")
-@click.option("--auto",
-              required=False,
-              default=False, is_flag=True,
-              help="Auto-find mode, adds bugs based on --group")
+@click.option("--mode",
+              required=True,
+              type=click.Choice(['LIST', 'SWEEP', 'PAYLOAD']),
```
tbielawa: Style comments: For other multiple-choice stuff we usually use all lower case. Also we don't have many other help commands going over several lines like that. Maybe best to include the specifics in the larger paragraph doctext?

tbielawa: Using mode is cool though. Getting rid of `--auto` will be more intuitive I think.

elliott
```diff
@@ -348,11 +353,17 @@ advisory.
 @click.option("--id", type=int, metavar='BUGID',
               multiple=True, required=False,
               help="Bugzilla IDs to add, conflicts with --auto [MULTIPLE]")
+@click.option("--old",
+              required=False,
+              help="Older payload to compare against")
+@click.option("--new",
+              required=False,
```
tbielawa: I think your idea on slack is a good one. Some kind of `--between A B` thing to see what's in `B` but not `A`. Don't even need to include fancy `...`s in there. Just set the `nargs` to 2 (I'm pretty sure click does that).

elliott
tbielawa: What if `list` was the default? I think I'd like that.

elliott
```diff
@@ -335,10 +336,14 @@ advisory.
...
-              required=False,
-              default=False, is_flag=True,
-              help="Auto-find mode, adds bugs based on --group")
+@click.option("--mode",
+              required=True,
+              type=click.Choice(['LIST', 'SWEEP', 'PAYLOAD']),
+              help="""Mode to use to find bugs: 
+  LIST - Specify a list of BZ IDs to add
+  SWEEP - Find bugs based on --group
+  PAYLOAD - Find bugs fixed between two payloads. You must provide two payloads to compare --old and --new.
```
tbielawa: `payload` is a kind of strange name for this. `mode` implies a verb to me. Sweeping is a verb, listing is a verb, but payload is a noun. Maybe call it `diff` ?

elliott
```diff
@@ -348,11 +353,17 @@ advisory.
...
+              help="Older payload to compare against")
+@click.option("--new",
+              required=False,
+              help="Newer payload to compare agaist")
 @click.option("--flag", metavar='FLAG',
               required=False, multiple=True,
               help="Optional flag to apply to found bugs [MULTIPLE]")
 @pass_runtime
-def find_bugs(runtime, advisory, default_advisory_type, auto, status, id, flag):
+def find_bugs(runtime, advisory, default_advisory_type, mode, status, id, old, new, flag):
```
tbielawa: Putting all those comments together, we'd have examples like this:

*Bugs between two payloads:* 
`elliott --group-openshift-4.1 find-bugs --mode diff --between 4.0.0-0.1 4.0.0-0.1` 

*List 4.1 bugs that would be attached (implicit `mode=list`):*
`elliott --group-openshift-4.1 find-bugs`

*Sweep bugs into 123456:*
`elliott --group-openshift-4.1 find-bugs --mode sweep --add 123456`

*Sweep bugs into default advisory:*
`elliott --group-openshift-4.1 find-bugs --mode sweep --use-default-advisory rpm`

Once you finalize this then some of those other examples will want to be looked at again.

elliottlib/openshiftclient.py
```diff
@@ -0,0 +1,24 @@
...
+# ours
+from elliottlib.exceptions import ElliottFatalError
+
+
+def get_bug_list(working_dir, old, new):
+    bug_list = []
+    try:
+        bug_list = check_output(['oc', 'adm', 'release', 'info',
+                                '-o', 'name', # only output BZ IDs
+                                '--bugs={}/origin'.format(working_dir), # clone origin to working dir
```
tbielawa: Does working dir get wiped away every time? We may want to think about adding that as a parameter in `~/.config/elliott/settings.yaml` so we can have it in a persistent spot? You will know better though since you're closer to this than I am right now.

elliott
```diff
@@ -388,30 +386,30 @@ advisory with the --add option.
...
 \b
-    $ elliott advisory:find-bugs --mode LIST --id 8675309 --id 7001337 --add 123456
+    $ elliott advisory:find-bugs --mode list --id 8675309 --id 7001337 --add 123456
 
     Automatically find bugs for openshift-4.1 and attach them to the
     rpm advisory defined in ocp-build-data:
 
 \b
-    $ elliott --group=openshift-4.1 --mode SWEEP --use-default-advisory rpm
+    $ elliott --group=openshift-4.1 --mode sweep --use-default-advisory rpm
```
tbielawa: plz add a real example for `diff` so it's easy to copy/pasta

tbielawa:
Liking the changes. Too bad about that merge conflict though.

elliottlib/openshiftclient.py
smunilla: I think it already is something you can put in there. I will double check.

elliott
```diff
@@ -361,13 +367,16 @@ described below:
 
     Note: Using --id without --add is basically pointless
 
-AUTOMATIC: For this use-case the --group option MUST be provided. The
+SWEEP: For this use-case the --group option MUST be provided. The
 --group automatically determines the correct target-releases to search
 for MODIFIED bugs in.
 
-MANUAL: The --group option is not required if you are specifying bugs
+LIST: The --group option is not required if you are specifying bugs
```
tbielawa: Maybe add words like `In LIST mode you must provide a list of IDs to attach with the --id option`

elliott
```diff
@@ -348,11 +350,15 @@ advisory.
 @click.option("--id", metavar='BUGID', default=None,
               multiple=True, required=False,
               help="Bugzilla IDs to add, conflicts with --auto [MULTIPLE]")
```
tbielawa: `--auto` doesn't exist anymore. Need to update `help` text for `--id`

elliott
```diff
@@ -335,10 +336,11 @@ advisory.
...
               help="Use the default value from ocp-build-data for ADVISORY_TYPE [image, rpm, security]")
-@click.option("--auto",
-              required=False,
-              default=False, is_flag=True,
-              help="Auto-find mode, adds bugs based on --group")
+@click.option("--mode",
+              required=True,
+              type=click.Choice(['list', 'sweep', 'diff']),
+              default='list',
+              help='Mode to use to find bugs')
```
tbielawa: Add to `help`, "list mode requires providing bugs with --id option"

elliott
```diff
@@ -377,34 +386,34 @@ advisory with the --add option.
...
-    $ elliott advisory:find-bugs --id 8675309 --id 7001337 --add 123456
+    $ elliott advisory:find-bugs --mode list --id 8675309 --id 7001337 --add 123456
 
     Automatically find bugs for openshift-4.1 and attach them to the
     rpm advisory defined in ocp-build-data:
 
 \b
-    $ elliott --group=openshift-4.1 --auto --use-default-advisory rpm
+    $ elliott --group=openshift-4.1 --mode sweep --use-default-advisory rpm
 """
```
tbielawa: Last thing. At the bottom of the docstring could you include a note like "Deprecation notice: In 0.1.11 the --auto mode was removed. Use --mode sweep now for equivalent functionality"

smunilla:
Superseded by #41

PR #55, opened 2019-10-11
[ART-1106] Revert "elliott find options fix"

thiagoalessio:
This reverts commit 3f68739a87b222aa14b7700cb41f42652121e77a.

https://jira.coreos.com/browse/ART-1047 states that one, and only one option out of `--kind`, `--from-diff` and `--build` should be allowed.

But the described use case does need 2 of those options to work `--kind` and `--build`).

Given that, I'd say the premise of ART-1047 is wrong, so we should just revert that commit.

### DIFF
--- a/elliott
+++ b/elliott
@@ -657,11 +657,8 @@ PRESENT advisory. Here are some examples:
     $ elliott --group openshift-3.6 find-builds -k rpm -b megafrobber-1.0.1-2.el7 -b 93170
 """
 
-    if (from_diff and kind) or (from_diff and builds) or (builds and kind):
-        raise ElliottFatalError("Use only one of --kind or --build or --from-diff.")
-
-    if not builds and not from_diff and not kind:
-        raise ElliottFatalError("Must use one of --kind or --build or --from-diff.")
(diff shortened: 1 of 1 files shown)


elliott
```diff
@@ -657,11 +657,8 @@ PRESENT advisory. Here are some examples:
     $ elliott --group openshift-3.6 find-builds -k rpm -b megafrobber-1.0.1-2.el7 -b 93170
 """
 
-    if (from_diff and kind) or (from_diff and builds) or (builds and kind):
-        raise ElliottFatalError("Use only one of --kind or --build or --from-diff.")
-
-    if not builds and not from_diff and not kind:
-        raise ElliottFatalError("Must use one of --kind or --build or --from-diff.")
```
sosiouxme: I think we still want this check for the original spirit of the reverted fix. We do need to specify one of these or we get a confusing error.

Additionally, IIRC if we specify `--build` we still need `--kind` or the code gets confused. So a condition to ensure that as well?

thiagoalessio:
Closing in favor of https://github.com/openshift/elliott/pull/57

PR #56, opened 2019-10-11
[ART-1091] Filter already attached image builds

thiagoalessio:
Make behavior consistent with `find-builds --kind rpm`.
Related ticket: https://jira.coreos.com/browse/ART-1091


sosiouxme:
Yes, I think it is that simple. I tried it, it works. Thanks!

(*Sigh* for code coverage)

PR #59, opened 2019-10-14
Move elliott 'create_placeholder' && 'list' commands into separated folder

shiywang:
Our `elliott.py` file contains more than 1500 line of code which is not really easy to read, I suggest we split it into different files into cli folder gradually like the subcommand @vfreex did on source tarball


vfreex:
I think some commands have been moved to the cli submodule by Luke. Please check and rebase.

PR #61, opened 2019-10-15
ART-1021 Fix the version unmatched

shiywang:
https://jira.coreos.com/projects/ART/issues/ART-1021

### DIFF
--- a/elliottlib/errata.py
+++ b/elliottlib/errata.py
@@ -275,6 +275,21 @@ def add_comment(advisory_id, comment):
                          data=data)
 
 
+def get_metadata_comments_json(advisory_id):
+    comments = get_comments(advisory_id)
+    metadata_json_list = []
+    # they come out in (mostly) reverse order, start at the beginning
+    for c in reversed(comments):
+        try:
+            metadata = json.loads(c['attributes']['text'])
+        except Exception:
+            pass
+        else:
+            if 'release' in metadata and 'kind' in metadata and 'impetus' in metadata:
+                metadata_json_list.append(metadata)
+    return metadata_json_list
+
+
 def get_comments(advisory_id):
     """5.2.10.2. GET /api/v1/comments?filter[key]=value
 
--- a/elliottlib/errata.py
+++ b/elliottlib/errata.py
@@ -275,21 +275,6 @@ def add_comment(advisory_id, comment):
                          data=data)
 
 
-def get_metadata_comments_json(advisory_id):
-    comments = get_comments(advisory_id)
-    metadata_json_list = []
-    # they come out in (mostly) reverse order, start at the beginning
-    for c in reversed(comments):
-        try:
-            metadata = json.loads(c['attributes']['text'])
-        except Exception:
-            pass
-        else:
-            if 'release' in metadata and 'kind' in metadata and 'impetus' in metadata:
-                metadata_json_list.append(metadata)
-    return metadata_json_list
-
-
 def get_comments(advisory_id):
     """5.2.10.2. GET /api/v1/comments?filter[key]=value
 
--- a/elliottlib/util.py
+++ b/elliottlib/util.py
@@ -251,8 +251,4 @@ def override_product_version(pv, branch):
 
 def get_release_version(pv):
     """ there are two kind of format of product_version: OSE-4.1-RHEL-8 RHEL-7-OSE-4.1 """
-    res = pv.split("OSE-", 1)[1]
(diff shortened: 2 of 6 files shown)


sosiouxme:
It certainly is pretty useful. Trying it against 4.1 today where I knew there were some updated RPMs hiding out. Before this fix:
```
$ elliott -g openshift-4.1 find-builds --kind rpm
...
The following 1 builds may be attached to an advisory:
 atomic-enterprise-service-catalog-4.1.21-201910171648.git.1.86d6b19.el7
```

After this fix:
```
$ elliott -g openshift-4.1 find-builds --kind rpm
The following 12 builds may be attached to an advisory:
 apb-2.0.3-2.el7
 atomic-enterprise-service-catalog-4.1.21-201910171648.git.1.86d6b19.el7
 containernetworking-plugins-0.8.1-4.el7
 golang-github-openshift-prometheus-alert-buffer-0-3.gitceca8c1.el7
 golang-github-prometheus-promu-0-5.git85ceabc.el7
 jenkins-2-plugins-4.1.1571057286-1.el7
 jenkins-2.176.3.1571056173-1.el7
 mediawiki-1.27.7-1.el7
 openshift-4.1.21-201910150726.git.0.27e6c28.el7
 openshift-ansible-4.1.21-201910150726.git.162.285bc08.el7
 openshift-eventrouter-0.2-3.gited73fb6.el7
 openshift-external-storage-0.0.2-8.gitd3c94f0.el7
```

So this is great. My concern was that many of these are already attached to the same advisory. I'm not certain if this is a problem, but it seems fishy; shouldn't they have been filtered out? What I don't want to happen is that RPMs get re-attached and have to go through RPMdiff again, not quite sure if that's what happens here.

sosiouxme:
Yeah; this is definitely a problem. Even after all of the RPMs are attached to two different advisories in 4.1, find-builds wants to add them all. That would be bad since signed-compose always adds everything elliott finds to main rpm advisory.

shiywang:
@sosiouxme this is weird, I assume this line we prevent it to do that.
```
# We only want builds not attached to an existing open advisory
        if b.attached_to_open_erratum:
``` 
do you know this is because of the problem of our tool or I misunderstanding this code ?

sosiouxme:
> @sosiouxme this is weird, I assume this line we prevent it to do that.
> 
> ```
> # We only want builds not attached to an existing open advisory
>         if b.attached_to_open_erratum:
> ```

Yes, that line is to prevent a build being considered viable for addition, if it's already been attached to an open erratum. What I thought your change was doing was to refine that test such that it is true only if the build is attached to an open erratum _with the same release_. Being attached to an erratum for some other release should be ignored.

What I saw when I ran it was that it was just including all unshipped builds regardless of whether they're attached to an open erratum.

sosiouxme:
Some unit tests might help ensure things are doing what they're supposed to :)

thiagoalessio:
> Some unit tests might help ensure things are doing what they're supposed to :)

+1

shiywang:
yeah, I got want you mean here, I updated one line code in my last commit to reflect my change, that definitely was a sloppy mistaken we could have avoided if we have tests.

I will start working on tests.

tbielawa:
I don't have a reproducer for this but it seems to work still!
/lgtm

elliottlib/errata.py
```diff
@@ -274,6 +274,21 @@ def add_comment(advisory_id, comment):
                          data=data)
 
 
+def get_metadata_comments_json(advisory_id):
```
sosiouxme: rebase and remove this, it's already merged in master

elliottlib/util.py
```diff
@@ -247,3 +247,12 @@ def override_product_version(pv, branch):
             return 'OSE-{}-RHEL-{}'.format(rb[1], rb[3])
         return 'RHEL-{}-OSE-{}'.format(rb[3], rb[1])
     return pv
+
+
+def get_release_version(pv):
+    """ there are two kind of format of product_version: OSE-4.1-RHEL-8 RHEL-7-OSE-4.1 """
+    res = pv.split("OSE-", 1)[1]
```
sosiouxme: this... works, but just seems really clunky. how about:
```suggestion
    return re.search(r'OSE-(\d+\.\d+)', pv).groups()[0]
```

shiywang: Yeah, regex is better

sosiouxme:
Yes! Thanks.

PR #62, opened 2019-10-15
ART-1055: tarball sources includes lookaside-cache sources from dist-git

vfreex:
When use `elliott tarball-sources create`, it also fetches lookaside-cache sources.

## For reviewers:

I used `eliott --debug tarball-sources create --component=openshift-clients --out-dir=out/ --force 43535` to create a source tarball for `openshift-clients` build that is attached to advisory `43535`.
By extracting the content of created tarball, you can see another tarball inside, which is the source stored in dist-git lookaside cache (defined in `sources` at dist-git repo root.)

### DIFF
--- a/elliottlib/cli/tarball_sources_cli.py
+++ b/elliottlib/cli/tarball_sources_cli.py
@@ -34,9 +34,10 @@ def tarball_sources_cli(ctx):
               default="hierarchical", show_default=True,
               help="Layout of output directory.")
 @click.option("--component", "components", multiple=True, help="Koji/Brew component names or build NVRs to filter on. Can be specified multiple times.")
-@click.option("-f", "--force", is_flag=True, help="Force overwrite existing files.")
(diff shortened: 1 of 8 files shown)


vfreex:
The CI error seems to be resulted from missing some native packages. Note: This PR introduces a new dependency `rpkg`, `fedpkg`'s core library.

vfreex:
I think the CI error was caused by failing to install `pycurl` due to `libcurl-devel` is not present in the test env.

tbielawa:
I'm not certain how to test this, can you give me some example commands and what I should look for when I run them?

vfreex:
@tbielawa I used `eliott --debug tarball-sources create --component=openshift-clients --out-dir=out/ --force 43535` to create a source tarball for `openshift-clients` build that is attached to advisory `43535`.

sosiouxme:
Deferring this until we actually need the feature

PR #60, opened 2019-10-15
adding find-builds suport attach multiple images with different pv

shiywang:
Depends on https://github.com/red-hat-storage/errata-tool/pull/180
Here's my test [result](https://errata.devel.redhat.com/advisory/47278/builds)

PR #64, opened 2019-10-17
ART-1062 adding find-builds suport attach multiple images with different pv

shiywang:
an alternative solution for https://github.com/openshift/elliott/pull/60 without touching errata-tool

### DIFF
--- a/elliottlib/brew.py
+++ b/elliottlib/brew.py
@@ -358,6 +358,7 @@ class BrewTaggedImageBuilds(object):
     builds. Ensures the result set is formatted correctly for this
     build type.
     """
+
     def __init__(self, tag):
         self.tag = tag
         self.builds = set([])
@@ -389,6 +390,7 @@ class BrewTaggedRPMBuilds(object):
     builds. Ensures the result set is formatted correctly for this
     build type.
     """
+
     def __init__(self, tag):
         self.tag = tag
         self.builds = set([])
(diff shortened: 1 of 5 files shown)


tbielawa:
How can I test this? Example commands would be useful

sosiouxme:
Seems to work. I tried it out on real data (after removing two test subjects) like so:
```
🐚 elliott -g openshift-4.2 find-builds --kind image 
...
The following 2 builds may be attached to an advisory:
 ironic-inspector-container-v4.2.1-201910171309
 jenkins-agent-maven-35-rhel7-container-v4.2.1-201910171309

🐚 elliott -g openshift-4.2 find-builds --kind image --use-default-advisory image
...
Attached build(s) successfully:
 ironic-inspector-container-v4.2.1-201910171309
 jenkins-agent-maven-35-rhel7-container-v4.2.1-201910171309
```

elliottlib/model.py
```diff
@@ -1,5 +1,5 @@
 
-class ModelException(StandardError):
+class ModelException(Exception):
```
sosiouxme: ... why?

shiywang: oh, this change was made by `autopep8` I use it to pass CI
the exception hierarchy tree is sth like this:
```
BaseException
 +-- SystemExit
 +-- KeyboardInterrupt
 +-- GeneratorExit
 +-- Exception
      +-- StopIteration
      +-- StandardError
```
So I guess `autopep8` decided to switch to `Exception` to help us catch more exceptions but base on our code is not that robust, I can change it back.

shiywang:
@sosiouxme @tbielawa  I guess for adding new commands or new options like `--from-diff` we should definitely add examples, this change it more insensible because I didn't introduce images and either bring confusing into the existing behavior of `elliott find-builds`

PR #65, opened 2019-10-17
Set correct permissions on rcm-guest for tarball sources

vfreex:
Per https://projects.engineering.redhat.com/browse/RCM-68294,
it would be more convenient for RCM if the file permissions
of tarball sources include read for other (o+r).

This PR uses `shutils.copyfile` rather than `shutils.copy2` to honor the umask for creating local tarballs,
and changes the instructions of sending to rcm-guest to explicitly set `r` for files and `rx` for directories.

### DIFF
--- a/elliottlib/cli/tarball_sources_cli.py
+++ b/elliottlib/cli/tarball_sources_cli.py
@@ -136,7 +136,7 @@ Use --force to add new tarball sources to an existing directory.".format(os.path
 
                 LOGGER.debug("Copying {} to {}...".format(
                     temp_tarball_path, tarball_abspath))
-                shutil.copy2(temp_tarball_path, tarball_abspath)
+                shutil.copyfile(temp_tarball_path, tarball_abspath)  # `shutil.copyfile` uses default umask
                 tarball_sources_list.append(tarball_abspath)
                 util.green_print(
                     "Created tarball source {}.".format(tarball_abspath))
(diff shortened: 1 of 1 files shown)


shiywang:
@vfreex how do I know this `shutils.copyfile` make the file (604) I suppose is the permission RCM wants

elliottlib/cli/tarball_sources_cli.py
```diff
@@ -136,7 +136,7 @@ def create(ctx, advisories, out_dir, out_layout, components, force):
 
                 LOGGER.debug("Copying {} to {}...".format(
                     temp_tarball_path, tarball_abspath))
-                shutil.copy2(temp_tarball_path, tarball_abspath)
+                shutil.copyfile(temp_tarball_path, tarball_abspath)  # `shutil.copyfile` uses default umask
```
shiywang: @vfreex how do I know this shutils.copyfile make the file (604) I suppose is the permission RCM wants

vfreex:
The reason I use default umask here is that should be the default behavior for creating new files, and the file mode should be `0644` by default unless changed.
To ensure we have `o+r` mode on rcm-guest, I think we can also update the instructional message for `rsync` to explicitly set the permission bit.

vfreex:
@shiywang PR updated, PTAL

elliottlib/cli/tarball_sources_cli.py
vfreex: Updated the rsync instruction

vfreex:
Also updated release management docs.

sosiouxme:
LGTM too :)

PR #68, opened 2019-10-18
Update OWNERS

shiywang:


thiagoalessio:
@shiywang nice, maybe you could take the opportunity to remove:
```
  - ahaile
  - smunilla
```
I believe `openshift-bot` keeps suggesting them as reviewers because of their presence in this file.

PR #69, opened 2019-10-18
remove dead code

shiywang:
Fixing https://gitlab.cee.redhat.com/openshift-art/tools/elliott/merge_requests/78/diffs#note_875105

PR #80, opened 2019-11-04
Refactor unit test imports for Python 3 migration

vfreex:
1. Use absolute imports.
2. Removes logging hacks from `tests/test_brew.py` (Not sure what is does and it causes troubles with `pytest`).
3. Rename `test_image.py` to `test_imagecfg.py` to be consistent with #69.

### DIFF
--- a/tests/test_assertion.py
+++ b/tests/test_assertion.py
@@ -3,7 +3,7 @@ Test the task related functions for the OpenShift Image/RPM Build Tool
 """
 import unittest
 
-import assertion
+from elliottlib import assertion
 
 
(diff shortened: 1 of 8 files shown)


vfreex:
@jupierce It seems to me that the `setUp()` and `tearDown()` methods in `TestBrew` can be safely removed. Can I remove it to make my local `pytest` working?

thiagoalessio:
/lgtm

just fix the conflicts and we are good to merge.

vfreex:
@thiagoalessio Thanks.

Issue #84, opened 2019-11-13
Python 3 support


yazug:
In starting to look at elliott I noticed a number of possible issues with python3 support in the code base.  Below are a few that I saw so far.

- tox missing py3 target (available in the python-3 branch)
- setup.py is missing test deps
- print() issue
- urlparse error
- unicode
- absolute vs relative imports
- iteritems
- dict values() is view not a list in python3
- contextlib.next was replace by ExitStack  (in a unittest)
- unittests and elliottlib.assertions.FileNotFoundError and ChildProcessError are not working in py3

I did see pull request #82 started looking at this

yazug:
Looks like there is topic branch of python-3 with some work

vfreex:
Fixed by #85 and #104.

PR #85, opened 2019-11-13
python 3 - compatiblity fixes

yazug:
split out a couple of other patches and have a long patch set of different python 2 vs 3 fixes/changes
tox -e py3 now works and passes locally

Most of these should be py2 backward compatible but some changes may impact runtime usage of the tool, more testing is required before anything can be said about whether it works or not.

Related: issue #84

### DIFF
--- a/elliottlib/cli/__main__.py
+++ b/elliottlib/cli/__main__.py
@@ -829,7 +829,7 @@ unsigned builds.
         all_signed = False
         # `errata_builds` is a dict with brew tags as keys, values are
         # lists of builds on the advisory with that tag
-        for k, v in e.errata_builds.iteritems():
+        for k, v in e.errata_builds.items():
             all_builds = all_builds.union(set(v))
         green_prefix("Fetching initial states: ")
         click.echo("{} builds to check".format(len(all_builds)))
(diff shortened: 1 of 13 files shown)


vfreex:
Thank you for helping with this. Yes we are currently in the progress of Python 3 migration. I will do a review in a few days.

vfreex:
Hi @yazug, could you do a rebase on top of `python-3` branch?

yazug:
> Hi @yazug, could you do a rebase on top of `python-3` branch?

done, updated and fixed up a couple of new instances of keys()/items() as lists issues

vfreex:
After looking at the PR, I think it is safe to merge. What this does are basically:

1. replaces old APIs with new ones:
    - `iteritems` to `items` for `dict`s.
    - explicitly converts the return value of `dict.keys()` and `dict.values()` to `list`.
    - import wrapper functions with `future.standard_library.install_aliases`
2. consolidate string types using `builtins.str`

I would like to give a LGTM. @thiagoalessio @sosiouxme @shiywang Please also take a look. I plan to merge it if we get another LGTM.

/lgtm

PR #87, opened 2019-11-18
adding oc client into dev container

shiywang:


.devcontainer/dev.Dockerfile
```diff
@@ -29,6 +29,10 @@ RUN wget -O /etc/yum.repos.d/rcm-tools-fedora.repo https://download.devel.redhat
   && dnf install -y rhpkg \
   && dnf clean all
 
+#include latest oc client
+RUN wget -O /tmp/openshift-client-linux-4.2.4.tar.gz https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/openshift-client-linux-4.2.4.tar.gz \
```
vfreex: 1. You need to delete the tarball from /tmp after installation.
2. Would be better to have the version number parameterized.

.devcontainer/dev.Dockerfile
```diff
@@ -29,6 +29,10 @@ RUN wget -O /etc/yum.repos.d/rcm-tools-fedora.repo https://download.devel.redhat
...
+  && tar -C /usr/local/bin -xzf  /tmp/openshift-client-linux-4.2.4.tar.gz
```
vfreex: Use `tar -C /path -xzf foo.tar.gz file1 file2` to filter the binaries out.

shiywang:
@vfreex PTAL

PR #99, opened 2019-12-09
Pass debug option down

joepvd:


vfreex:
LGTM but it would be nice if the called function can be refactored to use logging rather than regular print.

/lgtm

tbielawa:
Why is the tester getting all those weird import errors? Flakes?

PR #108, opened 2020-01-08
[ART-1176] refactoring elliott find_builds_cli.py bug fixes and removing branch option and product_version

shiywang:
This is the same PR with https://github.com/openshift/elliott/pull/103, but that one had some problems with CI pip install yaml package.

PR #110, opened 2020-01-17
ART-1466: Fix Elliott fails to attach RPMs are tagged to released tag…

vfreex:
… out of order

Currently Elliott find unshipped RPMs using the following approach:
1. Get a list A of latest RPMs with candidate tag from Brew.
2. Get a list B of latest RPMs with base tag from Brew.
3. Subtract B from A.

The problem is some RPMs that are latest with the candidate tag are not considered latest with the base tag.
In such case, Elliott attempts to attach shipped RPMs but fails with the following exception:

```
$ elliott --group openshift-3.11 find-builds -k rpm  --use-default-advisory rpm
2020-01-17 06:35:26,363 INFO Data clone directory already exists, checking commit sha
2020-01-17 06:35:26,843 INFO https://github.com/openshift/ocp-build-data.git is already cloned and latest
2020-01-17 06:35:26,906 INFO Using branch from group.yml: rhaos-3.11-rhel-7
Default advisory detected: 50532
Generating list of rpms: Hold on a moment, fetching Brew builds
Refreshing for tag: rhaos-3.11-rhel-7
Refreshing for tag: rhaos-3.11-rhel-7-candidate
Gathering additional information: Brew buildinfo is required to continue
[**********************]
Refreshing for tag: rhaos-3.11-rhel-8-candidate
Refreshing for tag: rhaos-3.11-rhel-8
Gathering additional information: Brew buildinfo is required to continue
[]
[**********************]

Traceback (most recent call last):
  File "/home/nay/yuxzhu/.local/bin/elliott", line 11, in <module>
    sys.exit(main())
  File "/home/nay/yuxzhu/.local/lib/python3.6/site-packages/elliottlib/cli/__main__.py", line 902, in main
    cli(obj={})
  File "/home/nay/yuxzhu/.local/lib/python3.6/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/nay/yuxzhu/.local/lib/python3.6/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/nay/yuxzhu/.local/lib/python3.6/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/nay/yuxzhu/.local/lib/python3.6/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/nay/yuxzhu/.local/lib/python3.6/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/home/nay/yuxzhu/.local/lib/python3.6/site-packages/click/decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "/home/nay/yuxzhu/.local/lib/python3.6/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/home/nay/yuxzhu/.local/lib/python3.6/site-packages/elliottlib/cli/find_builds_cli.py", line 127, in find_builds_cli
    _attach_to_advisory(unshipped_builds, kind, advisory)
  File "/home/nay/yuxzhu/.local/lib/python3.6/site-packages/elliottlib/cli/find_builds_cli.py", line 249, in _attach_to_advisory
    file_types={build.nvr: [file_type] for build in builds}
  File "/home/nay/yuxzhu/.local/lib/python3.6/site-packages/errata_tool/erratum.py", line 695, in addBuilds
    return self.addBuildsDirect(buildlist, release, **kwargs)
  File "/home/nay/yuxzhu/.local/lib/python3.6/site-packages/errata_tool/erratum.py", line 674, in addBuildsDirect
    self._processResponse(r)
  File "/home/nay/yuxzhu/.local/lib/python3.6/site-packages/errata_tool/connector.py", line 256, in _processResponse
    raise ErrataException(err_msg)
errata_tool.exception.ErrataException: Erratum 50532: Unable to add build 'openshift-monitor-project-lifecycle-3.11.51-2.git.59.7b59e29.el7' which has already been released.
```

This PR filters shipped builds out of the to-be-attached list.

### DIFF
--- a/elliottlib/cli/find_builds_cli.py
+++ b/elliottlib/cli/find_builds_cli.py
@@ -98,7 +98,7 @@ PRESENT advisory. Here are some examples:
     # get the builds we want to add
     unshipped_nvrps = []
     if builds:
-        green_prefix('Fetching Brew tags...')
+        green_prefix('Fetching tags for builds...')
         unshipped_nvrps = _fetch_builds_by_nvr_or_id(builds, tag_pv_map)
     elif from_diff:
         green_print('Fetching changed images between payloads...')
@@ -114,7 +114,7 @@ PRESENT advisory. Here are some examples:
         lambda nvrp: elliottlib.brew.get_brew_build('{}-{}-{}'.format(nvrp[0], nvrp[1], nvrp[2]), nvrp[3], session=requests.Session())
     )
 
-    unshipped_builds = _attached_to_open_erratum_with_correct_pv(kind, results, elliottlib.errata)
+    unshipped_builds = _filter_out_inviable_builds(kind, results, elliottlib.errata)
 
     _json_dump(as_json, unshipped_builds, kind, tag_pv_map)
 
@@ -214,16 +214,14 @@ def _fetch_builds_by_kind_rpm(tag_pv_map):
     return rpm_tuple
 
 
-def _attached_to_open_erratum_with_correct_pv(kind, results, errata):
-    if kind != "image":
-        return results
+def _filter_out_inviable_builds(kind, results, errata):
     unshipped_builds = []
     # will probably end up loading the same errata and
     # its comments many times, which is pretty slow
     # so we cached the result.
     errata_version_cache = {}
     for b in results:
-        same_version_exist = False
+        in_same_version = False
         # We only want builds not attached to an existing open advisory
         if b.attached_to_open_erratum:
             for e in b.open_errata_id:
@@ -238,9 +236,9 @@ def _attached_to_open_erratum_with_correct_pv(kind, results, errata):
                     # but not much point in doing it). just looking at the first one is fine.
                     errata_version_cache[e] = metadata_comments_json[0]['release']
                 if errata_version_cache[e] == get_release_version(b.product_version):
-                    same_version_exist = True
+                    in_same_version = True
                     break
-        if not same_version_exist or not b.attached_to_open_erratum:
+        if not (in_same_version and (b.attached_to_open_erratum or b.attached_to_shipped_erratum)):
             unshipped_builds.append(b)
     return unshipped_builds
 
--- a/elliottlib/cli/find_builds_cli.py
+++ b/elliottlib/cli/find_builds_cli.py
@@ -216,29 +216,26 @@ def _fetch_builds_by_kind_rpm(tag_pv_map):
 
 def _filter_out_inviable_builds(kind, results, errata):
     unshipped_builds = []
-    # will probably end up loading the same errata and
-    # its comments many times, which is pretty slow
-    # so we cached the result.
-    errata_version_cache = {}
+    errata_version_cache = {}  # avoid reloading the same errata for multiple builds
     for b in results:
+        # check if build is attached to any existing advisory for this version
         in_same_version = False
-        # We only want builds not attached to an existing open advisory
-        if b.attached_to_open_erratum:
-            for e in b.open_errata_id:
-                if not errata_version_cache.get(e):
-                    metadata_comments_json = errata.get_metadata_comments_json(e)
-                    if not metadata_comments_json:
-                        # Does not contain ART metadata, skip it
-                        red_print("Errata {} Does not contain ART metadata\n".format(e))
-                        continue
-                    # it's possible for an advisory to have multiple metadata comments,
-                    # though not very useful (there's a command for adding them,
-                    # but not much point in doing it). just looking at the first one is fine.
-                    errata_version_cache[e] = metadata_comments_json[0]['release']
-                if errata_version_cache[e] == get_release_version(b.product_version):
-                    in_same_version = True
-                    break
-        if not (in_same_version and (b.attached_to_open_erratum or b.attached_to_shipped_erratum)):
+        for eid in [e['id'] for e in b.all_errata]:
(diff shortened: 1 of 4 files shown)


elliottlib/cli/find_builds_cli.py
```diff
@@ -202,9 +202,7 @@ def _fetch_builds_by_kind_rpm(tag_pv_map):
     return rpm_tuple
 
 
-def _attached_to_open_erratum_with_correct_pv(kind, results, errata):
-    if kind != "image":
```
shiywang: This might be a potential problem IIRC since this filter with erratum thing used to be "only images", last time I was trying to remove this caused some problems.

elliottlib/cli/find_builds_cli.py
```diff
@@ -228,7 +226,7 @@ def _attached_to_open_erratum_with_correct_pv(kind, results, errata):
                 if errata_version_cache[e] == get_release_version(b.product_version):
                     same_version_exist = True
                     break
-        if not same_version_exist or not b.attached_to_open_erratum:
+        if (not same_version_exist or not b.attached_to_open_erratum) and not b.attached_to_shipped_erratum:
```
shiywang: according to De Morgan's laws, this could be reduced to 
```
if not (same_version_exist and b.attached_to_open_erratum or b.attached_to_shipped_erratum):
```
but for readability maybe like
```
if not ( same_version_exist and b.attached_to_open_erratum) and not b.attached_to_shipped_erratum:
```

elliottlib/cli/find_builds_cli.py
vfreex: Currently what this function does is filtering builds that has been shipped or other advisories with the same version (by reading the ART comment) out of the list to attach. I don't think there will be an issue with it. But better to confirm with @sosiouxme

elliottlib/cli/find_builds_cli.py
vfreex: I prefer the first one `if not (same_version_exist and b.attached_to_open_erratum or b.attached_to_shipped_erratum):`

elliottlib/cli/find_builds_cli.py
shiywang: Yeah, it was just my vague memories of having a problem on this part, maybe it's gone now. @sosiouxme do you know what was the reason why we use this errata filter only for images, not RPMs as well?

joepvd: I cannot judge whether removing this test will break an edge condition. It is however a cheap test. If we were to treat it as an `assert`, it'd probably more easy to live with it.

I'd say to undo the removal of `_attached_to_open_erratum_with_correct_pv` in this PR, so we can get it merged.

Maybe this function can log something when it takes an action so we can find  circumstances if we want to?

elliottlib/cli/find_builds_cli.py
joepvd: Any could go for me

vfreex: Updated.

vfreex:
@shiywang @joepvd I have clearly explained with this PR what will be filtered out when attaching a build. If you guys think it is unreasonable, then this bug will never be fixed.

elliottlib/cli/find_builds_cli.py
vfreex: @joepvd The point of this PR is not attaching shipped release RPMs to an advisory. If we remove this check, this PR will be meaningless.

sosiouxme: really sorry for not looking at this sooner.

@shiywang this condition is a recent addition of yours from https://github.com/openshift/elliott/commit/df8be4a9fa3d9af9b3b30af1dc3383972aa7076a#diff-0e3c3290b54e518e2da0eb3dc26df8ceR212
(part of https://github.com/openshift/elliott/pull/108 recently merged).

I don't think it makes sense (I also don't think the original name for this made sense, it was removing all the things _not_ attached to an open erratum, so @vfreex 's name change here seems good). we don't want to re-attach what's already been attached/shipped for the same version, whether it's an image or rpm. this _may_ have been what [pulled in already-attached rpms on 4.3.0](https://issues.redhat.com/browse/OCPPLAN-2622?focusedCommentId=13947211&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13947211) (though it's hard to tell which thing went wrong there).

elliottlib/cli/find_builds_cli.py
sosiouxme: can i suggest that the same treatment above for `if b.attached_to_open_erratum:` be applied also `if b.attached_to_shipped_erratum:` and feed into `same_version_exist` (perhaps rename that as `in_same_version`)?

then i think this becomes:
```suggestion
        if not in_same_version:
```
... because the only way that can ever be set is if it was already attached to an advisory for the same version.

vfreex:
@shiywang @joepvd @sosiouxme I have addressed the issues. PTAL. Thanks.

sosiouxme:
I added a commit of my own invention after spending way too long trying to wrap my head around logic. Going to merge now!

PR #116, opened 2020-02-18
[ART-1528 partial] add new cli command create candidate

shiywang:
I tried this afternoon it works with my personal access token 
https://github.com/openshift/cincinnati-graph-data/pull/55

So currently we need one `github access_token` and one `slack_api_token` and `ssh-keys`


shiywang:
@vfreex Sounds good to me, I am not opinionated on name picking, so `request-cincy-edge` if no objections since I think currently we are not going to add any drawback PR features contrast with `add`

Also as you guys can see here for this command to work I need at least github access token, ssh-pub and priv key any instructions on How could I get those or way to workaround?
@sosiouxme @thiagoalessio

thiagoalessio:
Relates to https://github.com/openshift/aos-cd-jobs/pull/2126

shiywang:
@sosiouxme @vfreex @thiagoalessio I closed this since https://github.com/openshift/aos-cd-jobs/pull/2153 has a better way to do it.

PR #124, opened 2020-04-17
Add y-stream trackers tool

sfowl:
The `y-stream-trackers.py` script logic can probably be integrated into elliott. This is PR is intended as a demonstration of the logic for determining if a y-stream tracker should be fixed in an RHSA.


shiywang:
LGTM, I think maybe we can integrate it into our create RHSA logic? @vfreex

thiagoalessio:
Thanks a lot @sfowl :heart: :yellow_heart: :blue_heart: :purple_heart:

PR #126, opened 2020-04-24
make some change to change_state to reflect more accurate error messages

shiywang:
```
[dev@1f06086f7f33 elliott]$ elliott change-state  --state QE --advisory 53516 
ElliottFatalError("Error: Could not change 'SHIPPED_LIVE' advisory 53516, group.yml is probably pointing at old one")
```

wrap up https://issues.redhat.com/browse/ART-1066

### DIFF
--- a/elliottlib/cli/change_state_cli.py
+++ b/elliottlib/cli/change_state_cli.py
@@ -1,0 +1,107 @@
+from __future__ import absolute_import, print_function, unicode_literals
+
+from elliottlib import logutil, Runtime
+from elliottlib.cli.common import cli, use_default_advisory_option, find_default_advisory
+from elliottlib.exceptions import ElliottFatalError
+from elliottlib.util import green_prefix, green_print
+
+from errata_tool import Erratum, ErrataException
+import click
+
+LOGGER = logutil.getLogger(__name__)
+
+pass_runtime = click.make_pass_decorator(Runtime)
+
+#
+# Set advisory state
+# change-state
+#
+@cli.command("change-state", short_help="Change ADVISORY state")
+@click.option("--state", '-s', required=True,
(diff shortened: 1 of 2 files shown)


sosiouxme:
would be nice as two commits, one for moving the source, and the second for the actual change... is it just the errmsg that changed?

elliottlib/cli/change_state_cli.py
```diff
@@ -0,0 +1,98 @@
...
+        prefix = ""
+
+    try:
+        e = Erratum(errata_id=advisory)
+
+        if e.errata_state == state:
+            green_prefix("{}No change to make: ".format(prefix))
+            click.echo("Target state is same as current state")
+            return
+        elif e.errata_state == 'SHIPPED_LIVE':
```
sosiouxme, line 78: This is not the only immutable state; there are several. Really, we can only change the state if it's in NEW_FILES or QE so if you want to check this I'd check the state is one of those. But frankly, passing through the ET error was fine.

elliottlib/cli/change_state_cli.py
```diff
@@ -0,0 +1,98 @@
...
+            raise ElliottFatalError("Error: Could not change 'SHIPPED_LIVE' advisory {advs}, group.yml is probably pointing at old one".format(advs=advisory))
```
sosiouxme, line 79: wouldn't say anything about group.yml unless `default_advisory_type` was set. they could have mistyped the number by hand and then this would be very confusing.

shiywang:
```
[dev@1f06086f7f33 elliott]$ elliott change-state  --state QE --advisory 53516 
ElliottFatalError("Error: we can only change the state if it's in NEW_FILES or QE, current state is SHIPPED_LIVE")
```

vfreex:
Basically LGTM if you split this into 2 commits.

PR #142, opened 2020-07-21
ART-997: Sweep (or don't sweep) bugs into correct advisory

vfreex:
3.y: Sweep all bugs into the rpm advisory.
4.y: Sweep optional operator bugs into extras advisory and other bugs into main image advisory.

@sosiouxme JIRA ART-997 doesn't describe which components/github repos we should ignore when sweeping bugs. For simplicity, the operator-related bugs are determined by a hardcoded list of bug components. Please have a look anyway.

### DIFF
--- a/elliottlib/cli/__main__.py
+++ b/elliottlib/cli/__main__.py
@@ -172,7 +172,7 @@ advisory with the --add option.
         raise click.BadParameter("If using mode=payload, you must provide two payloads to compare")
 
     if sum(map(bool, [advisory, default_advisory_type, into_default_advisories])) > 1:
-        raise click.BadParameter("Use only one of --use-default-advisory, --add, or --into_default_advisories")
+        raise click.BadParameter("Use only one of --use-default-advisory, --add, or --into-default-advisories")
 
     runtime.initialize()
     bz_data = runtime.gitdata.load_data(key='bugzilla').data
@@ -253,7 +253,7 @@ advisory with the --add option.
     # Otherwise we don't need to sweep bugs at all.
     if not (into_default_advisories or default_advisory_type):
         return
-    impetus_bugs = {}  # key is rpm impetus ("rpm", "image", "extras"), value is a set of bug IDs.
+    impetus_bugs = {}  # key is impetus ("rpm", "image", "extras"), value is a set of bug IDs.
     # @lmeyer: simple and stupid would still be keeping the logic in python, possibly with config flags for branched logic. until that logic becomes too ugly to keep in python, i suppose..
     if major_version < 4:  # for 3.x, all bugs should go to the rpm advisory
         impetus_bugs["rpm"] = {bug.id for bug in bugs}
--- a/elliottlib/cli/__main__.py
+++ b/elliottlib/cli/__main__.py
@@ -192,7 +192,40 @@ advisory with the --add option.
         bug_id_strings = elliottlib.openshiftclient.get_bug_list(runtime.working_dir, from_diff[0], from_diff[1])
         bugs = [bzapi.getbug(i) for i in bug_id_strings]
 
-    green_prefix("Found {} bugs:".format(len(bugs)))
+    # Some bugs should goes to CPaaS so we should ignore them
+    m = re.match(r"rhaos-(\d+).(\d+)", runtime.branch)  # extract OpenShift version from the branch name. there should be a better way...
+    if not m:
+        raise ElliottFatalError(f"Unable to determine OpenShift version from branch name {runtime.branch}.")
+    major_version = int(m[1])
+    minor_version = int(m[2])
+
+    def _filter_bugs(bugs):  # returns a list of bugs that should be processed
+        r = []
+        ignored_repos = set()  # GitHub repos that should be ignored
+        if major_version == 4 and minor_version == 5:
+            # per https://issues.redhat.com/browse/ART-997: these repos should have their release-4.5 branches ignored by ART:
+            ignored_repos = {
+                "https://github.com/openshift/aws-ebs-csi-driver",
+                "https://github.com/openshift/aws-ebs-csi-driver-operator",
+                "https://github.com/openshift/cloud-provider-openstack",
+                "https://github.com/openshift/csi-driver-nfs",
+                "https://github.com/openshift/csi-driver-manila-operator"
+            }
+        for bug in bugs:
+            external_links = [ext["type"]["full_url"].replace("%id%", ext["ext_bz_bug_id"]) for ext in bug.external_bugs]  # https://github.com/python-bugzilla/python-bugzilla/blob/7aa70edcfea9b524cd8ac51a891b6395ca40dc87/bugzilla/_cli.py#L750
+            public_links = [runtime.get_public_upstream(url)[0] for url in external_links]  # translate openshift-priv org to openshift org when comparing to filter (i.e. prow may link to a PR on the private org).
+            # if a bug has 1 or more public_links, we should ignore the bug if ALL of the public_links are ANY of `ignored_repos`
+            if public_links and all(map(lambda url: any(map(lambda repo: url != repo and url.startswith(repo), ignored_repos)), public_links)):
+                continue
+            r.append(bug)
+        return r
+
+    if len(id) == 0:  # unless --id is given, we should ignore bugs that don't belong to ART. e.g. some bugs should go to CPaaS
+        filtered_bugs = _filter_bugs(bugs)
+        green_prefix(f"Found {len(filtered_bugs)} bugs ({len(bugs) - len(filtered_bugs)} ignored):")
+        bugs = filtered_bugs
+    else:
+        green_prefix("Found {} bugs:".format(len(bugs)))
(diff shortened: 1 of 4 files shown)


elliottlib/cli/__main__.py
```diff
@@ -200,12 +208,37 @@ def find_bugs(runtime, advisory, default_advisory_type, mode, status, id, from_d
...
+    impetus_bugs = {}  # key is rpm impetus ("rpm", "image", "extras"), value is a set of bug IDs.
+    m = re.match(r"rhaos-(\d+).(\d+)", runtime.branch)  # extract OpenShift version from the branch name. there should be a better way...
+    if not m:
+        raise ElliottFatalError(f"Unable to determine OpenShift version from branch name {runtime.branch}.")
+    # @lmeyer: simple and stupid would still be keeping the logic in python, possibly with config flags for branched logic. until that logic becomes too ugly to keep in python, i suppose..
+    major_version = int(m[1])
+    if major_version < 4:  # for 3.x, all bugs should go to the rpm advisory
+        impetus_bugs["rpm"] = {bug.id for bug in bugs}
+    else:  # for 4.x, optional operators bugs should be swept to the "extras" advisory, while other bugs should be swept to "image" advisory.
+        # a way to identify operator-related bugs is by its "Component" value. temporarily hardcode here until we need to move it to ocp-build-data.
```
wking, line 231: you can look in `oc adm release info --commits "${RELEASE_PULLSPEC}"` to get a list of repos and commits, and then you can look into their `OWNERS` to get a Bugzilla component.  For example:

```console
$ oc adm release info -o json quay.io/openshift-release-dev/ocp-release:4.5.2-x86_64 | jq -r '.references.spec.tags[].annotations | .["io.openshift.build.source-location"] as $repo | ($repo | sub("https://github.com/"; "")) as $org_repo | .["io.openshift.build.commit.id"] as $commit | select($commit | length > 0) | "https://raw.githubusercontent.com/" + $org_repo + "/" + $commit + "/OWNERS"' | sort | uniq | while read URI; do COMPONENT="$(curl -s "${URI}" | yaml2json | jq -r .component)"; echo "${URI} -> ${COMPONENT}"; done
https://raw.githubusercontent.com/openshift/aws-pod-identity-webhook/2251f7e2fc9194aa3a57cd9dc1de4a71e9367df5/OWNERS -> Cloud Credential Operator
https://raw.githubusercontent.com/openshift/baremetal-operator/e276e9f64e3c49abbd13267914bfa7fef1fa3a9b/OWNERS -> Bare Metal Hardware Provisioning
https://raw.githubusercontent.com/openshift/baremetal-runtimecfg/855ed6b0022c8c9557138e58514bb4301f4043f5/OWNERS -> null
https://raw.githubusercontent.com/openshift/builder/bb6e41a1e23a61e070274f778d86d4211bdb41ff/OWNERS -> null
https://raw.githubusercontent.com/openshift/cloud-credential-operator/863098a4b783c2194dac62eddfc55fc44c8f7ae4/OWNERS -> Cloud Credential Operator
...
```

That's not going to be perfect, but if we fill in some remaining `component` values where that gives `null` today, it should be pretty close for identifying bugs which affect a release image directly.  Things like machine-os-content that pull in content via RPMs would be more work.  Everything else would be `extras`, right?

sosiouxme: if the OWNERS file specifies BZ info then it's reflected in labels on each image built from it. but that's not terribly useful here, as we aren't looking at builds here but rather bugs, and trying to determine if they pertain to the payload or to extras.

we don't yet have a reliable link between the bug and the build it corresponds to. if we established that... it might be easier to build a mapping between build usage and related bugs.

elliottlib/cli/__main__.py
```diff
@@ -163,8 +171,8 @@ def find_bugs(runtime, advisory, default_advisory_type, mode, status, id, from_d
     if mode == 'payload' and not len(from_diff) == 2:
         raise click.BadParameter("If using mode=payload, you must provide two payloads to compare")
 
-    if advisory and default_advisory_type:
-        raise click.BadParameter("Use only one of --use-default-advisory or --add")
+    if sum(map(bool, [advisory, default_advisory_type, into_default_advisories])) > 1:
+        raise click.BadParameter("Use only one of --use-default-advisory, --add, or --into_default_advisories")
```
sosiouxme, line 175: tiny correction...
```suggestion
        raise click.BadParameter("Use only one of --use-default-advisory, --add, or --into-default-advisories")
```

elliottlib/cli/__main__.py
```diff
@@ -200,12 +208,37 @@ def find_bugs(runtime, advisory, default_advisory_type, mode, status, id, from_d
...
 
     if len(flag) > 0:
-        for bug in bug_ids:
+        for bug in bugs:
             for f in flag:
                 bug.updateflags({f: "+"})
 
-    if advisory is not False:
-        elliottlib.errata.add_bugs_with_retry(advisory, [bug.id for bug in bug_ids], False)
+    if advisory is not False and not default_advisory_type:  # `--add ADVISORY_NUMBER` should respect the user's wish and attach all available bugs to whatever advisory is specified.
```
sosiouxme, line 215: specifying both of these is already rejected above... i think it adds only confusion to assert the same here, and would be clearer left as
```suggestion
    if advisory:  # `--add ADVISORY_NUMBER` should respect the user's wish and attach all available bugs to whatever advisory is specified.
```

elliottlib/cli/__main__.py
```diff
@@ -200,12 +208,37 @@ def find_bugs(runtime, advisory, default_advisory_type, mode, status, id, from_d
...
+        elliottlib.errata.add_bugs_with_retry(advisory, [bug.id for bug in bugs], False)
+        return
+
+    # if --use-default-advisory or --into-default-advisories is given, we need to determine which bugs should be swept into which advisory
+    if not into_default_advisories and not default_advisory_type:
+        return
+    impetus_bugs = {}  # key is rpm impetus ("rpm", "image", "extras"), value is a set of bug IDs.
```
sosiouxme, line 222:
```suggestion
    impetus_bugs = {}  # key is impetus ("rpm", "image", "extras"), value is a set of bug IDs.
```

elliottlib/cli/__main__.py
```diff
@@ -200,12 +208,37 @@ def find_bugs(runtime, advisory, default_advisory_type, mode, status, id, from_d
...
+    m = re.match(r"rhaos-(\d+).(\d+)", runtime.branch)  # extract OpenShift version from the branch name. there should be a better way...
+    if not m:
+        raise ElliottFatalError(f"Unable to determine OpenShift version from branch name {runtime.branch}.")
+    # @lmeyer: simple and stupid would still be keeping the logic in python, possibly with config flags for branched logic. until that logic becomes too ugly to keep in python, i suppose..
```
sosiouxme, line 226: not sure this comment is helpful :)

elliottlib/cli/__main__.py
```diff
@@ -200,12 +208,37 @@ def find_bugs(runtime, advisory, default_advisory_type, mode, status, id, from_d
...
             for f in flag:
                 bug.updateflags({f: "+"})
 
-    if advisory is not False:
-        elliottlib.errata.add_bugs_with_retry(advisory, [bug.id for bug in bug_ids], False)
+    if advisory is not False and not default_advisory_type:  # `--add ADVISORY_NUMBER` should respect the user's wish and attach all available bugs to whatever advisory is specified.
+        elliottlib.errata.add_bugs_with_retry(advisory, [bug.id for bug in bugs], False)
+        return
+
+    # if --use-default-advisory or --into-default-advisories is given, we need to determine which bugs should be swept into which advisory
```
sosiouxme, line 219: and if not, we don't need to sweep bugs at all... that may be worth saying here. or simply move this comment below the conditional.

elliottlib/cli/__main__.py
wking: [`oc adn release info --bugs ...`][1] is heavy, but will list bugs addressed between two releases.  Except for the first GA in a z-stream, because during feature development we don't require formal bug -> PR links.

[1]:https://github.com/openshift/oc/blob/b6f8f97e76dd72ee9a7189701d548f16bd38de27/pkg/cli/admin/release/info.go#L93

elliottlib/cli/__main__.py
vfreex: `advisory` could be either set by user (with `--add`) or `--use-default-advisory` (https://github.com/openshift/elliott/pull/142/files#diff-13669dcc79bc03ff33da296c047bd280R182).

We will attach all bugs if that variable is set by `--add` not `--use-default-advisory`.

elliottlib/cli/__main__.py
vfreex: I've amended the comment.

vfreex:
The new push adds support for filtering by source repos and addresses other issues.

vfreex:
`get_public_upstream` and `convert_remote_git_to_https` are copied from Doozer without any changes.

elliottlib/cli/__main__.py
sosiouxme: i see; alright then!

PR #149, opened 2020-09-01
Rebuild tox environment after change requirements

joepvd:
Tox does not detect changes in requirements, and will fail with missing
imports.
[example](https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/art-tools/job/elliott/job/master/6/console)
[tox bug](https://github.com/tox-dev/tox/issues/149)

This PR looks at the files that have changed in the last 5 commits, and
if one of the files that might contain requirements have changed, will
add the argument `--recreate` to the `tox` invocation.

The magical number 5 was chosen, as it is conceivable that there are
multiple commits locally which contain the change, and the change in
dependencies is not detected.

Alternative approaches are:
- install a tox plugin
- always recreate in Jenkins

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -13,7 +13,10 @@ pipeline {
             steps {
                 script {
                     catchError(stageResult: 'FAILURE') {
-                        sh "tox > results.txt 2>&1"
+                        sh """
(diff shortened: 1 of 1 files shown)


thiagoalessio (approved):
/lgtm

PR #159, opened 2020-10-14
ART-2139: Fix elliott find-builds --remove doing nothing

vfreex:
This bug is not obvious but will happen if a build is untagged from Brew
after attaching to an advisory.

The reason is that the `--remove` functionality is added to the existing
sweeping builds logic, which needs to figure out the product version of
each specified build based on the Brew tag. However, if a build is
untagged, it does not map to any product versions so it will be filtered
out eventually.

Removing a build from advisory doesn't really require product versions.
This PR splits the removing build operation to a separate function and
adds more verifications to make things clear.


joepvd (approved):
/lgtm

sosiouxme (approved):

PR #162, opened 2020-10-26
[ART-2287] Create advisory-drop command

joepvd:
Advisories can only be dropped by their creators, or by program
management. This change implements the advisory-drop command, which
allows a bot user that creates advisories, to also drop them.


joepvd:
Have been playing on staging errata: https://errata.stage.engineering.redhat.com/advisory/61150

sosiouxme (approved):
:shipit:

PR #168, opened 2020-11-18
add new field 'for_release' to individual files

shiywang:


elliottlib/metadata.py
```diff
@@ -129,3 +129,7 @@ def get_component_name(self):
             component_name = self.config.distgit.component
 
         return component_name
+
+    @property
+    def is_release(self):
```
vfreex, line 134: Could you explain what `for_release` is used for?  Also the property name `is_release` seems to be more confusing than `for_release`.

vfreex (changes requested):

elliottlib/metadata.py
shiywang: here is the card: https://issues.redhat.com/browse/ART-2292
is_release just a name consistent with is_payload

PR #175, opened 2021-01-07
ART-2520 (2) Fix target_release fetched from bz instead of git data

thegreyd:
Earlier we were getting target_release out of [ocp-build-data's bugzilla.yml](https://github.com/openshift/elliott/blob/master/elliottlib/cli/attach_cve_flaws_cli.py#L59) instead we need to fetch it from the tracker bug(s). 

If there are conflicting target_release values found for tracker bugs, the script notifies and fails.


thiagoalessio (approved):
/lgtm

PR #176, opened 2021-02-03
ART-2592 Attach RPM CVE tracker bugs to RPM advisory

thegreyd:
**Summary**
- https://issues.redhat.com/browse/ART-2592
- This is a change to elliott find-bugs sweep functionality (see the full command below), to detect rpm cve tracker bugs and attach those to the rpm advisory
- The bugs are validated against brew builds attached to advisory. If builds are not found then bugs are not attached, so the ideal sweep workflow needed for this change would be sweep builds and then sweep bugs.

**Status**
- [X] Code changes
- ~~Some unit tests~~ - No unit tests exist for find bugs - would do a follow up story after current refactoring see https://github.com/openshift/elliott/pull/177
- [X] Prepare_release job changes if needed - https://github.com/openshift/aos-cd-jobs/pull/2531

**Testing:**
- `./elliott -g openshift-4.6 find-bugs --mode sweep --cve-trackers --into-default-advisories --check-builds --dry-run`
- `./elliott -g openshift-4.7 find-builds -k rpm` <- verify it runs after refactoring

### DIFF
--- a/elliottlib/brew.py
+++ b/elliottlib/brew.py
@@ -197,15 +197,12 @@ def find_unshipped_build_candidates(base_tag, kind='rpm', brew_session=koji.Clie
     :return: A set of build strings where each build is only tagged as
     a -candidate build
     """
-    shipped_builds_set = set()
+    shipped_builds_set = {b['nvr'] for b in brew_session.listTagged(tag=base_tag, latest=True, type=kind)}
+    candidate_builds_set = {b['nvr'] for b in brew_session.listTagged(tag='{}-candidate'.format(base_tag), latest=True,
+                                                             type=kind)}
     diff_builds = {}
-    candidate_builds = brew_session.listTagged(tag='{}-candidate'.format(base_tag), latest=True, type=kind)
-    for b in brew_session.listTagged(tag=base_tag, latest=True, type=kind):
-        shipped_builds_set.add(b['nvr'])
-
-    for b in candidate_builds:
-        if b['nvr'] not in shipped_builds_set:
-            diff_builds[b['nvr']] = b
+    for b in candidate_builds_set - shipped_builds_set:
+        diff_builds[b['nvr']] = b
 
     return diff_builds
 
--- a/elliottlib/brew.py
+++ b/elliottlib/brew.py
@@ -30,7 +30,8 @@ from requests_kerberos import HTTPKerberosAuth
 logger = logutil.getLogger(__name__)
 
 
-def get_tagged_builds(tags: Iterable[str], build_type: Optional[str], event: Optional[int], session: koji.ClientSession) -> List[Optional[List[Dict]]]:
+def get_tagged_builds(tags: Iterable[str], build_type: Optional[str], event: Optional[int],
+                      session: koji.ClientSession) -> List[Optional[List[Dict]]]:
     """ Get tagged builds for multiple Brew tags
 
     :param tags: List of tag names
@@ -44,7 +45,8 @@ def get_tagged_builds(tags: Iterable[str], build_type: Optional[str], event: Opt
     return [build for task in tasks for build in task.result]
 
 
-def get_latest_builds(tag_component_tuples: List[Tuple[str, str]], session: koji.ClientSession) -> List[Optional[List[Dict]]]:
+def get_latest_builds(tag_component_tuples: List[Tuple[str, str]], session: koji.ClientSession) -> List[
+    Optional[List[Dict]]]:
     """ Get latest builds for multiple Brew components
 
     :param tag_component_tuples: List of (tag, component_name) tuples
@@ -89,7 +91,8 @@ def wait_tasks(task_ids: Iterable[int], session: koji.ClientSession, sleep_secon
                 waiting_tasks.discard(task_id)  # remove from the wait list
         if waiting_tasks:
             if logger:
-                logger.debug(f"There are still {len(waiting_tasks)} tagging task(s) running. Will recheck in {sleep_seconds} seconds.")
+                logger.debug(
+                    f"There are still {len(waiting_tasks)} tagging task(s) running. Will recheck in {sleep_seconds} seconds.")
             time.sleep(sleep_seconds)
 
 
@@ -198,11 +201,12 @@ def find_unshipped_build_candidates(base_tag, kind='rpm', brew_session=koji.Clie
     a -candidate build
     """
     shipped_builds_set = {b['nvr'] for b in brew_session.listTagged(tag=base_tag, latest=True, type=kind)}
-    candidate_builds_set = {b['nvr'] for b in brew_session.listTagged(tag='{}-candidate'.format(base_tag),
-                                                                      latest=True, type=kind)}
+    candidate_builds = {b['nvr']: b for b in brew_session.listTagged(tag='f{base_tag}-candidate',
+                                                                         latest=True, type=kind)}
+    candidate_builds_set = candidate_builds.keys()
     diff_builds = {}
-    for b in candidate_builds_set - shipped_builds_set:
-        diff_builds[b['nvr']] = b
+    for nvr in candidate_builds_set - shipped_builds_set:
+        diff_builds[nvr] = candidate_builds[nvr]
 
     return diff_builds
 
--- a/elliottlib/brew.py
+++ b/elliottlib/brew.py
@@ -201,7 +201,7 @@ def find_unshipped_build_candidates(base_tag, kind='rpm', brew_session=koji.Clie
     a -candidate build
     """
     shipped_builds_set = {b['nvr'] for b in brew_session.listTagged(tag=base_tag, latest=True, type=kind)}
-    candidate_builds = {b['nvr']: b for b in brew_session.listTagged(tag='f{base_tag}-candidate',
+    candidate_builds = {b['nvr']: b for b in brew_session.listTagged(tag=f'{base_tag}-candidate',
                                                                          latest=True, type=kind)}
     candidate_builds_set = candidate_builds.keys()
     diff_builds = {}
--- a/elliottlib/brew.py
+++ b/elliottlib/brew.py
@@ -45,8 +45,7 @@ def get_tagged_builds(tags: Iterable[str], build_type: Optional[str], event: Opt
     return [build for task in tasks for build in task.result]
 
 
-def get_latest_builds(tag_component_tuples: List[Tuple[str, str]], session: koji.ClientSession) -> List[
-    Optional[List[Dict]]]:
+def get_latest_builds(tag_component_tuples: List[Tuple[str, str]], session: koji.ClientSession) -> List[Optional[List[Dict]]]:
     """ Get latest builds for multiple Brew components
 
     :param tag_component_tuples: List of (tag, component_name) tuples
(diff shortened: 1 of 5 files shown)


elliottlib/brew.py
```diff
@@ -197,15 +197,12 @@ def find_unshipped_build_candidates(base_tag, kind='rpm', brew_session=koji.Clie
...
     diff_builds = {}
-    candidate_builds = brew_session.listTagged(tag='{}-candidate'.format(base_tag), latest=True, type=kind)
-    for b in brew_session.listTagged(tag=base_tag, latest=True, type=kind):
-        shipped_builds_set.add(b['nvr'])
-
-    for b in candidate_builds:
-        if b['nvr'] not in shipped_builds_set:
-            diff_builds[b['nvr']] = b
+    for b in candidate_builds_set - shipped_builds_set:
+        diff_builds[b['nvr']] = b
```
vfreex, line 205: `b` is of type `str`, not `dict` anymore.
```python
diff_builds = candidate_builds_set - shipped_builds_set
```

elliottlib/cli/find_bugs_cli.py
```diff
@@ -220,16 +220,52 @@ def _filter_bugs(bugs):  # returns a list of bugs that should be processed
...
+    # @lmeyer: simple and stupid would still be keeping the logic in python,
+    # possibly with config flags for branched logic.
+    # until that logic becomes too ugly to keep in python, i suppose..
     if major_version < 4:  # for 3.x, all bugs should go to the rpm advisory
         impetus_bugs["rpm"] = set(bugs)
     else:  # for 4.x
-        # optional operators bugs should be swept to the "extras" advisory, while other bugs should be swept to "image" advisory.
-        # a way to identify operator-related bugs is by its "Component" value. temporarily hardcode here until we need to move it to ocp-build-data.
+        # sweep rpm cve trackers into "rpm" advisory
+        if mode == 'sweep' and cve_trackers:
```
vfreex, line 238: If `cve_trackers` is false, we still need to sweep "normal" bugs here. Just CVE trackers are not swept.

elliottlib/cli/find_bugs_cli.py
```diff
@@ -220,16 +220,52 @@ def _filter_bugs(bugs):  # returns a list of bugs that should be processed
...
+            rpm_bugs = bzutil.get_valid_rpm_cves(bugs)
+            green_prefix("RPM CVEs found: ")
+            click.echo([b.id for b in rpm_bugs])
+
+            # check if corresponding brew builds are attached to advisory
+            attached_builds = errata.get_advisory_nvrs(runtime.group_config.advisories["rpm"])
```
vfreex, line 244: Does this mean we need to assume builds are already swept before running a bug sweep?
I get a feeling this will become a gotcha if we adopt this assumption.
There are multiple ways to distinguish if a component is an image or rpm. All brew container images are ending with `-container` suffix.

elliottlib/cli/find_bugs_cli.py
```diff
@@ -220,16 +220,52 @@ def _filter_bugs(bugs):  # returns a list of bugs that should be processed
...
+            packages = attached_builds.keys()
+            failed = []
+            for bug, package_name in rpm_bugs.items():
+                if package_name not in packages:
+                    failed.append((bug.id, package_name))
```
vfreex, line 249: CVEs not belonging to rpm advisory should go to the main image or the extras advisory, i guess.

elliottlib/cli/find_bugs_cli.py
```diff
@@ -220,16 +220,52 @@ def _filter_bugs(bugs):  # returns a list of bugs that should be processed
...
+
+        # optional operators bugs should be swept to the "extras" advisory
+        # a way to identify operator-related bugs is by its "Component" value.
+        # temporarily hardcode here until we need to move it to ocp-build-data.
         extra_components = {"Logging", "Service Brokers", "Metering Operator", "Node Feature Discovery Operator"}  # we will probably find more
         impetus_bugs["extras"] = {b for b in bugs if b.component in extra_components}
-        impetus_bugs["image"] = {b for b in bugs if b.component not in extra_components}
+
+        # all other bugs should go into "image" advisory
+        impetus_bugs["image"] = set(bugs) - impetus_bugs["extras"] - impetus_bugs["rpm"]
```
vfreex, line 268: bugs that go to the rpm advisory should be excluded as well.

elliottlib/bzutil.py
```diff
@@ -290,6 +294,25 @@ def is_cve_tracker(bug_obj):
     return "SecurityTracking" in bug_obj.keywords and "Security" in bug_obj.keywords
 
 
+def get_valid_rpm_cves(bugs):
```
vfreex, line 297: I think you misunderstood the bug whiteboard here. IIUC ProdSec will label CVE trackers with their corresponding Brew component names, including both rpms and images. In Brew's convention, container image names are all ending with `-container`, while rpms are not. You can determine whether a CVE tracker should be attached to the rpm advisory or image advisory by checking its suffix.

vfreex (changes requested):
Hi @thegreyd , thank you for working on this.
I reviewed your PR, but it seems to me this is not correct due to misunderstanding the bug convention. Since you are new to those stuffs, could you discuss with the team (especially @sosiouxme ) about your idea first? Feel free to ask any questions.

elliottlib/cli/find_bugs_cli.py
thegreyd: I'm not sure what I'm missing here.. 

bugs attached to image = all bugs - (bugs attached to extras) - (bugs attached to rpm)

Or are you saying the `impetus_bugs["rpm"]` would not have all rpm advisory bugs in some case?

elliottlib/cli/find_bugs_cli.py
thegreyd: They would be at line 268..

elliottlib/cli/find_bugs_cli.py
thegreyd: Yes, this is a design choice I wasn't sure of myself, and I would prefer to keep `sweep bugs` and `sweep builds` independent. One alternative I looked at was
- get rpm cves and get the package names
- get unreleased nvrs from brew and search for package names in that (this is the sweep build logic)

But this would be pretty heavy, and if anything changes brew-wise between sweep bugs run and sweep build run, then things would be inconsistent. 

I think we should keep the validation optional, and by default attach bugs to the rpm advisory. If at sweep bugs time, more than 0 builds are attached and if package name is not found - then raise warning. If in this flow we notice that this actually doesn't work then we come back and make it not-optional. 

@sosiouxme thoughts?

elliottlib/cli/find_bugs_cli.py
thegreyd: I asked @sosiouxme about this part, specifically attaching normal (non-cve) bugs to rpm advisory and from what I understood is, we cannot rely for every bug to have the rpm identification logic (`component:` or `-container` if we're using that). I could be wrong, let me know and I'd be happy to change this logic.

elliottlib/bzutil.py
thegreyd: Thanks for this. I went with what's mentioned in the story, in case that's not correct or incomplete I can update it. @sosiouxme @smunilla

elliottlib/brew.py
thegreyd: :100: Thanks for catching this!

elliottlib/cli/find_bugs_cli.py
sosiouxme: As mentioned in above points, relying on the naming convention is probably better than coupling this to sweep order.

Re "validation optional" also remember that container image CVEs come in with whiteboard `component:openshift-golang-builder-container` instead of the full set of images actually affected; I think we need to tolerate that.

sosiouxme (commented):
What I noticed in my harried review...

elliottlib/bzutil.py
```diff
@@ -290,6 +294,25 @@ def is_cve_tracker(bug_obj):
...
+    An OCP rpm cve tracker has a whiteboard value "component:<component_name>"
+
+    :param bugs: list of bug objects
+    :returns: A dict of bug object as key and component name as value
+    """
+    marker = 'component:'
+    rpm_cves = {}
+    for b in bugs:
+        if is_cve_tracker(b) and marker in b.whiteboard:
+            tmp = b.whiteboard.split(marker)
```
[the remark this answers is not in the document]
thegreyd: +1

elliottlib/bzutil.py
thegreyd: Make sense!

elliottlib/brew.py
```diff
@@ -197,15 +197,12 @@ def find_unshipped_build_candidates(base_tag, kind='rpm', brew_session=koji.Clie
     :return: A set of build strings where each build is only tagged as
     a -candidate build
     """
-    shipped_builds_set = set()
+    shipped_builds_set = {b['nvr'] for b in brew_session.listTagged(tag=base_tag, latest=True, type=kind)}
+    candidate_builds_set = {b['nvr'] for b in brew_session.listTagged(tag='{}-candidate'.format(base_tag),
+                                                                      latest=True, type=kind)}
```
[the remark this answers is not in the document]
thegreyd: +1

elliottlib/cli/find_bugs_cli.py
sfowl: Validation is probably necessary at some stage. If naming convention alone is used, it could be possible for tracker bugs to be added to an advisory, without a corresponding build (this seems to happen a lot with the jenkins rpms, which aren't shipped release). It's painful to rectify this, the security role is need to remove security tracking bugs from advisories.

thegreyd: sfowl that is a valid case, may I suggest a new command `remove-orphan-bugs` ? This would be independent of both the commands, and would help in manual workflow as well.

But to your point of having the validation step, rethinking about all possible options, the easiest seems for us to sweep builds and then sweep bugs validating with builds. Maybe a `--validate` flag to force validation, runs without it would not have that.

sfowl: If `remove-orphan-bugs` removes bugs from advisories I think that would require the user to have security permissions in ET, which I don't think ART members have. We want members to be able to prepare correct advisories independently.

If validation is optional then I imagine it's possible for bugs to be attached to the wrong advisory, which then requires non-ART members to get involved, which we want to avoid.

thegreyd: Thanks for clarifying that. @sosiouxme I've added the `--check-builds` flag to enforce validation, it will skip attaching bugs if builds are not found. Ready for review

thegreyd: Thinking about it, a run without this flag would result in rpm bugs still being attached to image advisory. Either we can skip adding those.. or have the flag always on.?

thegreyd: It's interesting that the cost of adding bugs is low but removing is high.. so we really have to get it right the first time or skip to allow for manual addition

sosiouxme (approved):

PR #177, opened 2021-02-09
ART-2445 Add find-bugs mode to check for blocker bugs

thegreyd:
## Links
- https://issues.redhat.com/browse/ART-2445
- Previous PR https://github.com/openshift/aos-cd-jobs/pull/2519

## Summary:
This PR adds a new mode to find-bugs 'blocker', which looks up blocker bugs for a group. 
- By default cve trackers are included
- Also a new flag `--exclude-status` which lets you exclude bug status

## TODO:
- [X] The list of component filters in story is different from what we have in bugzilla.yml, ~~make sure that bugzilla.yml is updated with that~~ After discussion with @sosiouxme we agreed to use the "default" component filter list in bugzilla.yml without any modification, since that is what ART uses in other places. The list could be changed/modified in the future accordingly.
- [x] prepare-release and promote job PR https://github.com/openshift/aos-cd-jobs/pull/2534

## Test
- `./elliott -g openshift-4.7 find-bugs --mode blocker --report`
- `./elliott -g openshift-4.7 find-bugs --mode blocker --exclude-status ON_QA --report`

### DIFF
--- a/elliottlib/bzutil.py
+++ b/elliottlib/bzutil.py
@@ -223,7 +223,8 @@ def search_for_bugs(bz_data, status, search_filter='default', flag=None, filter_
     :param bz_data: The Bugzilla data dump we got from our bugzilla.yaml file
     :param status: The status(es) of bugs to search for
     :param search_filter: Which search filter from bz_data to use if multiple are specified
-    :param flag: A specific flag to filter out bugs
+    :param flag: The Bugzilla flag (string) of bugs to search for. Flags are similar to status but are categorically
+    different. https://bugzilla.readthedocs.io/en/latest/using/understanding.html#flags
     :param filter_out_security_bugs: Boolean on whether to filter out bugs tagged with the SecurityTracking keyword.
 
     :return: A list of Bug objects
--- a/elliottlib/bzutil.py
+++ b/elliottlib/bzutil.py
@@ -217,18 +217,19 @@ def create_placeholder(bz_data, kind):
     return newbug
 
 
-def search_for_bugs(bz_data, status, search_filter='default', filter_out_security_bugs=True, verbose=False):
+def search_for_bugs(bz_data, status, search_filter='default', flag=None, filter_out_security_bugs=True, verbose=False):
     """Search the provided target_release's for bugs in the specified states
 
     :param bz_data: The Bugzilla data dump we got from our bugzilla.yaml file
     :param status: The status(es) of bugs to search for
     :param search_filter: Which search filter from bz_data to use if multiple are specified
+    :param flag: A specific flag to filter out bugs
     :param filter_out_security_bugs: Boolean on whether to filter out bugs tagged with the SecurityTracking keyword.
 
     :return: A list of Bug objects
     """
     bzapi = get_bzapi(bz_data)
-    query_url = _construct_query_url(bz_data, status, search_filter)
+    query_url = _construct_query_url(bz_data, status, search_filter, flag=flag)
 
     fields = ['id', 'status', 'summary', 'creation_time', 'cf_pm_score', 'component', 'external_bugs']
 
@@ -327,7 +328,7 @@ def get_bzapi(bz_data, interactive_login=False):
     return bzapi
 
 
-def _construct_query_url(bz_data, status, search_filter='default'):
+def _construct_query_url(bz_data, status, search_filter='default', flag=None):
     query_url = SearchURL(bz_data)
 
     if bz_data.get('filter'):
@@ -343,6 +344,9 @@ def _construct_query_url(bz_data, status, search_filter='default'):
 
     for r in bz_data.get('target_release', []):
         query_url.addTargetRelease(r)
+
+    if flag:
+        query_url.addFlagFilter(flag, "substring")
 
     return query_url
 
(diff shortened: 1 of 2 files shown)


elliottlib/cli/find_bugs_cli.py
```diff
@@ -21,20 +21,27 @@
...
+              default=None,
+              required=False,
+              type=click.Choice(constants.VALID_BUG_STATES),
+              help="Exclude bugs of this status. Useful when using default statuses")
 @click.option("--id", metavar='BUGID', default=None,
               multiple=True, required=False,
               help="Bugzilla IDs to add, required for LIST mode.")
 @click.option("--cve-trackers",
               required=False,
+              default=None,
```
thegreyd, line 44: I wasn't sure if there was a way to set this to False if the default was True. Like `--cve-trackers=False`, if not the case I'll revert this change. Part of the exploration of "different modes have different defaults, but if a user explicitly sets it then use that".

elliottlib/cli/find_bugs_cli.py
```diff
@@ -21,20 +21,27 @@
...
               help='Mode to use to find bugs')
 @click.option("--status", 'status',
               multiple=True,
+              default=None,
               required=False,
-              default=['MODIFIED', 'VERIFIED', 'ON_QA'],
               type=click.Choice(constants.VALID_BUG_STATES),
-              help="Status of the bugs")
+              help="Include bugs of this status")
+@click.option("--exclude-status", 'exclude_status',
```
thegreyd, line 33: This was added so we can use blocker in prepare-release job with `--exclude-status ON_QA`

elliottlib/cli/find_bugs_cli.py
sosiouxme: i'm pretty sure `--cve-trackers=False` works fine

elliottlib/cli/find_bugs_cli.py
```diff
@@ -123,49 +137,74 @@ def find_bugs_cli(runtime, advisory, default_advisory_type, mode, check_builds,
...
-        raise click.BadParameter("--mode=qe does not operate on an advisory. Do not specify any of `--use-default-advisory`, `--add`, or `--into-default-advisories`")
+    if mode in ['qe', 'blocker'] and advisory_attach_flags > 0:
+        raise click.BadParameter("Mode does not operate on an advisory. Do not specify any of "
+                                 "`--use-default-advisory`, `--add`, or `--into-default-advisories`")
 
     runtime.initialize()
     bz_data = runtime.gitdata.load_data(key='bugzilla').data
     bzapi = bzutil.get_bzapi(bz_data)
 
+    # Some bugs should goes to CPaaS so we should ignore them
```
sosiouxme, line 166: this comment seems out of place. and wherever it might go, I think I'd prefer something along the lines of "filter out bugs ART does not manage"

elliottlib/cli/find_bugs_cli.py
```diff
@@ -123,49 +137,74 @@ def find_bugs_cli(runtime, advisory, default_advisory_type, mode, check_builds,
...
 
-    if mode == 'sweep' or mode == 'qe':
-        if mode == 'qe':
-            status = ['MODIFIED']
+    if mode in ['sweep', 'qe', 'blocker']:
+        if not cve_trackers:
+            if mode == 'blocker':
+                cve_trackers = True
+            else:
+                cve_trackers = False
```
sosiouxme, line 182: Seems a bit verbose, how about:
```suggestion
            cve_trackers = mode == 'blocker'
```

elliottlib/cli/find_bugs_cli.py
```diff
@@ -123,49 +137,74 @@ def find_bugs_cli(runtime, advisory, default_advisory_type, mode, check_builds,
...
+
+        if not status:
```
sosiouxme, line 184: how about a comment?
```suggestion
        if not status:  # use default status filter according to mode
```

sosiouxme (commented):
alright, i have nits, but works well and LGTM aside from the nits (which I'll leave to your discretion)

elliottlib/cli/find_bugs_cli.py
thegreyd: Not sure about that, I prefer the first one since the branching makes it very clear that cve_trackers value varies by mode. The value could be bool, or string doesn't matter, the structure would be: `if not param: if mode == 'X': param = Y`

thegreyd: Also this is soon to be refactored into individual function for each mode, so we would get rid of this anyway :)

elliottlib/bzutil.py
```diff
@@ -217,18 +217,19 @@ def create_placeholder(bz_data, kind):
...
 
 
-def search_for_bugs(bz_data, status, search_filter='default', filter_out_security_bugs=True, verbose=False):
+def search_for_bugs(bz_data, status, search_filter='default', flag=None, filter_out_security_bugs=True, verbose=False):
     """Search the provided target_release's for bugs in the specified states
 
     :param bz_data: The Bugzilla data dump we got from our bugzilla.yaml file
     :param status: The status(es) of bugs to search for
     :param search_filter: Which search filter from bz_data to use if multiple are specified
+    :param flag: A specific flag to filter out bugs
```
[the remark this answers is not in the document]
thegreyd: Added
```python
:param flag: The Bugzilla flag (string) of bugs to search for. Flags are similar to status but are categorically
    different. https://bugzilla.readthedocs.io/en/latest/using/understanding.html#flags
```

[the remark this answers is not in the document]
thegreyd: But I'm not sure what else we can rename the param, since it's called flag everywhere in bz.

thegreyd:
> alright, i have nits, but works well and LGTM aside from the nits (which I'll leave to your discretion)

Thank you for the review! I've addressed the comments.

PR #178, opened 2021-02-18
ART-2692 Elliott first-fix version parsing fixes

thegreyd:
## Summary
- https://issues.redhat.com/browse/ART-2692

## Test
`./elliott -g openshift-4.7 attach-cve-flaws --use-default-advisory rpm --dry-run`

### DIFF
--- a/elliottlib/attach_cve_flaws.py
+++ b/elliottlib/attach_cve_flaws.py
@@ -1,7 +1,4 @@
-import requests
-import ssl
-from elliottlib import constants, bzutil, errata, util
-from requests_kerberos import HTTPKerberosAuth
+from elliottlib import constants, errata, util
 
 
 def get_attached_tracker_bugs(bzapi, advisory_id):
@@ -36,16 +33,11 @@ def get_corresponding_flaw_bugs(bzapi, tracker_bugs):
     return [flaw_bug for flaw_bug in blocking_bugs if is_flaw_bug(flaw_bug)]
 
 
-def is_first_fix(bzapi, flaw_bug, current_target_release, tracker_ids_to_be_ignored=[]):
+def is_first_fix(bzapi, flaw_bug, tracker_ids_to_be_ignored=[]):
     other_flaw_trackers = bzapi.query(bzapi.build_query(
         product='OpenShift Container Platform',
         bug_id=[t for t in flaw_bug.depends_on if t not in tracker_ids_to_be_ignored],
     ))
-
-    def _filter_tracker(bug):
-        current_major_version = util.minor_version_tuple(current_target_release[0])[0]
-        bug_target_major_version = util.minor_version_tuple(bug.target_release[0])[0]
-        return bug_target_major_version == current_major_version
 
     def _already_fixed(bug):
         if bug.status == 'RELEASE_PENDING':
@@ -54,7 +46,7 @@ def is_first_fix(bzapi, flaw_bug, current_target_release, tracker_ids_to_be_igno
             return True
         return False
 
-    return not any([_already_fixed(t) for t in filter(_filter_tracker, other_flaw_trackers)])
+    return not any([_already_fixed(t) for t in other_flaw_trackers])
 
 
 def is_security_advisory(advisory):
--- a/elliottlib/util.py
+++ b/elliottlib/util.py
@@ -238,6 +238,33 @@ def parallel_results_with_progress(inputs, func, file=None):
     return results
 
 
+def get_target_release(bugs):
+    invalid_bugs = []
+    target_releases = set()
+    for bug in bugs:
+        # make sure it's a list with a valid str value
+        valid_target_rel = isinstance(bug.target_release, list) and len(bug.target_release) > 0 and \
+                           re.match(r'(\d+.\d+.[0|z])', bug.target_release[0])
+        if not valid_target_rel:
+            invalid_bugs.append(bug)
+        else:
+            target_releases.add(bug.target_release[0])
+
+    err = ''
+    if invalid_bugs:
+        err = f'bug.target_release should be a list with a string matching regex (\d+.\d+.[0|z])'
+        for b in invalid_bugs:
+            err += f'\n bug.id: {b.id}, bug.target_release: {b.target_release} '
+        return '', err
+
+    if len(target_releases) != 1:
+        err = f'Found different target_release values for tracker bugs: {target_releases}. ' \
+              'There should be only 1 target release for all bugs. Fix the offending bug(s) and try again.'
+        return '', err
+
+    return target_releases.pop(), ''
+
+
 def get_release_version(pv):
     """ there are two kind of format of product_version: OSE-4.1-RHEL-8 RHEL-7-OSE-4.1 RHEL-7-OSE-4.1-FOR-POWER-LE """
     return re.search(r'OSE-(\d+\.\d+)', pv).groups()[0]
--- a/elliottlib/cli/attach_cve_flaws_cli.py
+++ b/elliottlib/cli/attach_cve_flaws_cli.py
@@ -1,7 +1,6 @@
-import bugzilla
-import click
+import bugzilla, click, re
 from errata_tool import Erratum
-from elliottlib import Runtime, bzutil
+from elliottlib import util
 from elliottlib.cli.common import cli, use_default_advisory_option, find_default_advisory
 from elliottlib.attach_cve_flaws import get_attached_tracker_bugs, \
     get_corresponding_flaw_bugs, is_first_fix, is_security_advisory, \
@@ -51,12 +50,11 @@ def attach_cve_flaws_cli(runtime, advisory_id, noop, default_advisory_type):
     if len(attached_tracker_bugs) == 0:
         exit(0)
 
-    target_releases = set(bug.target_release[0] for bug in attached_tracker_bugs)
-    if len(target_releases) != 1:
-        runtime.logger.info('found different target_release for tracker bugs: {}'.format(target_releases))
+    # validate and get target_release
+    current_target_release, err = util.get_target_release(attached_tracker_bugs)
+    if err:
+        runtime.logger.error(err)
         exit(1)
-
-    current_target_release = target_releases.pop()
     runtime.logger.info('current_target_release: {}'.format(current_target_release))
 
     corresponding_flaw_bugs = get_corresponding_flaw_bugs(bzapi, attached_tracker_bugs)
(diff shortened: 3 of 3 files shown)


elliottlib/util.py
```diff
@@ -238,6 +238,33 @@ def parallel_results_with_progress(inputs, func, file=None):
     return results
 
 
+def get_target_release(bugs):
```
thegreyd, line 241: This function gets a list of bugs attached to an advisory, it then validates that the target release of all those bugs is the one single exact value example `4.7.0`, cannot be `4.7.z`. I assumed that was the right logic. Is that assumption correct?

sfowl: I think it's a bit more complicated than that I'm afraid.

Each flaw bug can have many trackers:
```
flaw1 -> tracker1
      -> tracker2
      -> tracker3
      -> tracker4
      -> tracker5           
```

Each tracker can have any value for it's target release:
```
flaw1 -> tracker1: 4.7.0
      -> tracker2: 4.7.0
      -> tracker3: 4.8.0
      -> tracker4: 4.7.z
      -> tracker5: 4.6.0
```

To work out if it's a first fix or not we need to look at the _component_:
```
flaw1 -> podman:  4.7.0
      -> cri-o:   4.7.0
      -> skopeo:  4.8.0
      -> skopeo:  4.7.z
      -> buildah: 4.6.0
```
For 3/4 components above, they only have one bug so that bug represents the first time that component will be fixed, i.e. it will be a first fix:

```
flaw1 -> podman:  4.7.0 (first fix)
      -> cri-o:   4.7.0 (first fix)
      -> skopeo:  4.8.0
      -> skopeo:  4.7.z
      -> buildah: 4.6.0 (first fix)
```
Only for skopeo do we need extra checks, because there is more than one tracker bug for skopeo. _If_ the 4.7.z skopeo bug has been fixed, only then will the 4.8.0 skopeo _not_ be a first fix (because it has been fixed already in z-stream). If the 4.7.z skopeo bug has not been fixed, then the 4.8.0 skopeo bug will be the first fix.

NB: z-stream bugs area always, always first fixes. We only need these first fix checks for .0 bugs, e.g. 4.7.0.

sosiouxme: TL;DR i think is that aside from

> We only need these first fix checks for .0 bugs, e.g. 4.7.0.

... 4.y.0 and 4.y.z are equivalent.

sosiouxme: that said, this was just about making sure all the bugs on the advisory are for the same target, not the trackers for a flaw bug.

yes, there should be a unique answer - 4.y.0 bugs should change to 4.y.z bugs if they don't make GA.

elliottlib/attach_cve_flaws.py
```diff
@@ -36,25 +33,20 @@ def get_corresponding_flaw_bugs(bzapi, tracker_bugs):
...
 
 
-def is_first_fix(bzapi, flaw_bug, current_target_release, tracker_ids_to_be_ignored=[]):
+def is_first_fix(bzapi, flaw_bug, tracker_ids_to_be_ignored=[]):
     other_flaw_trackers = bzapi.query(bzapi.build_query(
         product='OpenShift Container Platform',
         bug_id=[t for t in flaw_bug.depends_on if t not in tracker_ids_to_be_ignored],
     ))
 
-    def _filter_tracker(bug):
```
thegreyd, line 45: This only-major-version-match filter is removed in favor of a strict check of 1 exact target release value for all bugs

elliottlib/util.py
thegreyd: So it looks to me that we're trying to answer the question **Which (ocp) release is considered the "first-fix" release for the component X for the cve flaw Y?** <- let me know if this is accurate

This means that our first fix logic needs to be enhanced with what you describe. Right now we're checking if any tracker bug for a flaw bug has been fixed (or is being fixed), then it's not a first-fix, otherwise it is. The component checks is missing.

thegreyd: I think our definition is stricter, it just takes flaw and tracker release versions. So right now we'll be missing attaching flaw bugs

elliottlib/attach_cve_flaws.py
```diff
@@ -37,24 +34,37 @@ def get_corresponding_flaw_bugs(bzapi, tracker_bugs):
...
-    def _already_fixed(bug):
+    def already_fixed(bug):
         if bug.status == 'RELEASE_PENDING':
             return True
         if bug.status == 'CLOSED' and bug.resolution in ['ERRATA', 'CURRENTRELEASE', 'NEXTRELEASE']:
             return True
         return False
 
-    return not any([_already_fixed(t) for t in filter(_filter_tracker, other_flaw_trackers)])
+    # if any tracker bug for the flaw bug
```
thegreyd, line 60: I refactored this to make it **very clear** of what was happening for future developers working on this :)

elliottlib/util.py
sfowl: > So it looks to me that we're trying to answer the question Which (ocp) release is considered the "first-fix" release for the component X for the cve flaw Y? <- let me know if this is accurate

There can, and often are, many first fixes. I.e. z-stream fixes are always first fixes for that given stream. It's only possible for a .0 fix to be a non-first fix for the entire product, if it's already been fixed in an earlier z-stream. In that case, the .0 release would not be a security update for the customer.

.0 fixes are only fist fixes, if it has not yet been fixed in an earlier z-stream.

elliottlib/attach_cve_flaws.py
```diff
@@ -37,24 +34,37 @@ def get_corresponding_flaw_bugs(bzapi, tracker_bugs):
...
 
 def is_first_fix(bzapi, flaw_bug, current_target_release, tracker_ids_to_be_ignored=[]):
-    other_flaw_trackers = bzapi.query(bzapi.build_query(
+    """
+    Check if a flaw bug is considered a first-fix for a target release
+    """
+    # get all the tracker bugs for a flaw bug
+    # but only for OCP product
+    tracker_ids = [t for t in flaw_bug.depends_on if t not in tracker_ids_to_be_ignored]
+    tracker_bugs = bzapi.query(bzapi.build_query(
```
sfowl, line 43: I think you need to only be comparing against trackers for the same component.

elliottlib/attach_cve_flaws.py
sfowl: Need to mention components here. CVEs can affect multiple components, we need to check if a tracker represents a first fix for the given component, not the general CVE.

elliottlib/attach_cve_flaws.py
```diff
@@ -37,24 +34,37 @@ def get_corresponding_flaw_bugs(bzapi, tracker_bugs):
 
 
 def is_first_fix(bzapi, flaw_bug, current_target_release, tracker_ids_to_be_ignored=[]):
-    other_flaw_trackers = bzapi.query(bzapi.build_query(
+    """
+    Check if a flaw bug is considered a first-fix for a target release
+    """
+    # get all the tracker bugs for a flaw bug
+    # but only for OCP product
+    tracker_ids = [t for t in flaw_bug.depends_on if t not in tracker_ids_to_be_ignored]
```
sfowl, line 42: You might need to also filter on keywords: "Security" and "SecurityTracking"

elliottlib/util.py
thegreyd: thanks sam!

elliottlib/attach_cve_flaws.py
thegreyd: Thanks, since this is something new I'll write up a story for this

elliottlib/attach_cve_flaws.py
thegreyd: This will be fixed in upcoming story.

elliottlib/attach_cve_flaws.py
thegreyd: Added that check, thanks!

elliottlib/attach_cve_flaws.py
thegreyd: @sfowl Please review https://issues.redhat.com/browse/ART-2725

elliottlib/attach_cve_flaws.py
```diff
@@ -1,23 +1,24 @@
...
 
 
-def is_tracker_bug(bug):
-    return 'Security' in bug['keywords'] and 'SecurityTracking' in bug['keywords']
+def is_tracker_bug(bug=None, keywords=None):
+    if bug is None and keywords is None:
+        raise ValueError("must pass at least 1 param, either bug object or value of bug keywords")
+    if keywords is None:
+        keywords = bug.keywords
+    return 'Security' in keywords and 'SecurityTracking' in keywords
```
sosiouxme, line 21: to be honest this does not look simpler to me

elliottlib/util.py
```diff
@@ -238,6 +238,35 @@ def parallel_results_with_progress(inputs, func, file=None):
...
+        for b in invalid_bugs:
+            err += f'\n bug.id: {b.id}, bug.target_release: {b.target_release} '
+        return '', err
+
+    if len(target_releases) != 1:
+        err = f'Found different target_release values for tracker bugs: {target_releases}. ' \
+              'There should be only 1 target release for all bugs. Fix the offending bug(s) and try again.'
+        return '', err
+
+    return target_releases.pop(), ''
```
sosiouxme, line 267: this all feels very golang. wouldn't it be more pythonic if it raised `exceptions.ElliottFatalError` to be caught?

elliottlib/util.py
```diff
@@ -238,6 +238,35 @@ def parallel_results_with_progress(inputs, func, file=None):
     return results
 
 
+def get_target_release(bugs):
```
sosiouxme, line 241: Looks like a good subject for unit or functional tests

sosiouxme (approved):
These aren't objections - it works fine and could merge as-is. I leave it to your discretion if you want to make any changes.

elliottlib/attach_cve_flaws.py
thegreyd: yes it's a weird case of passing bug dict that comes from errata and trying to use the bug obj that comes from bugzilla.. There are ways to convert the dict into object type notations, like namedtuple etc and I wasn't sure if that was simpler. Hmm thinking more about it..

elliottlib/util.py
thegreyd: +1 I missed golang I guess.. but as you said not pythonic. I'll change it

elliottlib/util.py
thegreyd: +1 will add it

elliottlib/attach_cve_flaws.py
thegreyd: Changing the param to keywords only

thiagoalessio (approved):
/lgtm

PR #180, opened 2021-02-24
ART-2452 Fix list mode

thegreyd:
- https://issues.redhat.com/browse/ART-2452
- [duplicate] https://issues.redhat.com/browse/ART-2640

## Summary
List mode was broken as in the correct variables weren't being set, too many branching conditions in one huge function being one of the main problems which is hurting us here. I extracted out the mode into its own function. Wanted to keep the PR clean and limit the scope so I intentionally left out the other modes as is, for another PR.

## Test
- Remove placeholder bug 1931306 from rpm advisory https://errata.devel.redhat.com/advisory/69705
- Attach it `./elliott -g openshift-4.6 --debug find-bugs --mode list --id 1931306 --use-default-advisory rpm`
- Verify it worked

### DIFF
--- a/elliottlib/cli/find_bugs_cli.py
+++ b/elliottlib/cli/find_bugs_cli.py
@@ -255,29 +255,30 @@ advisory with the --add option.
             green_prefix("RPM CVEs found: ")
             click.echo(sorted(b.id for b in rpm_bugs))
 
-            # if --check-builds flag is set
-            # only attach bugs that have corresponding brew builds attached to rpm advisory
-            if check_builds:
-                click.echo("Validating bugs with builds attached to the rpm advisory")
-                attached_builds = errata.get_advisory_nvrs(runtime.group_config.advisories["rpm"])
-                packages = attached_builds.keys()
-                not_found = []
-                for bug, package_name in rpm_bugs.items():
-                    if package_name not in packages:
-                        not_found.append((bug.id, package_name))
-                    else:
-                        click.echo(f"Build found for #{bug.id}, {package_name}")
-                        impetus_bugs["rpm"].add(bug)
+            if rpm_bugs:
+                # if --check-builds flag is set
+                # only attach bugs that have corresponding brew builds attached to rpm advisory
+                if check_builds:
+                    click.echo("Validating bugs with builds attached to the rpm advisory")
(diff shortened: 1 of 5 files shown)


elliottlib/cli/find_bugs_cli.py
```diff
@@ -130,6 +130,9 @@ def find_bugs_cli(runtime, advisory, default_advisory_type, mode, check_builds,
     if mode == 'list' and len(id) == 0:
         raise click.BadParameter("When using mode=list, you must provide a list of bug IDs")
 
+    if mode == 'list' and into_default_advisories:
```
vfreex, line 133: Although this is rare but I think `list` mode can be used with `--into-default-advisories`. `list` just means "use specified bugs"

elliottlib/cli/find_bugs_cli.py
```diff
@@ -288,3 +256,63 @@ def _filter_bugs(bugs):  # returns a list of bugs that should be processed
...
+def mode_list(advisory, bugs, report, flags, noop):
+    green_prefix(f"Found {len(bugs)} bugs: ")
+    click.echo(", ".join(sorted(str(b.bug_id) for b in bugs)))
+    if report:
+        print_report(bugs)
+
+    if flags:
+        add_flags(bugs, flags)
+
+    errata.add_bugs_with_retry(advisory, bugs, noop=noop)
```
vfreex, line 317: `advisory` could be `None`.

vfreex (changes requested):
Thanks for addressing this issue and the refactor. Just a few minor issues, otherwise LGTM.

elliottlib/cli/find_bugs_cli.py
thegreyd: I think if not necessary we shouldn't support such a rare usecase (of attaching same bugs to multiple advisories). I checked with Luke, and we agreed on not having this option usable. Let me know if you still think we should have it

elliottlib/cli/find_bugs_cli.py
thegreyd: Thanks for catching this!

thiagoalessio (approved):
/lgtm

PR #182, opened 2021-03-16
[ART-2663] change query scope

Ximinhan:


elliottlib/bzutil.py
```diff
@@ -287,7 +288,7 @@ def is_cve_tracker(bug_obj):
     :param bug_obj: bug object
     :returns: True if the bug is a CVE tracker.
     """
-    return "SecurityTracking" in bug_obj.keywords and "Security" in bug_obj.keywords
+    return "SecurityTracking" in bug_obj.keywords or "Security" in bug_obj.keywords
```
vfreex, line 291: IIRC a tracker bug must have both `SecurityTracking` and `Security` keywords. A bug only having `Security` keyword could be a flaw bug (if it is assigned to `Security Response` product). @sfowl Could you provide more info?

vfreex (changes requested):

elliottlib/bzutil.py
sfowl: @vfreex you're understanding is correct.

PR #187, opened 2021-04-05
Sweep builds by brew event

thegreyd:
This is a change that will help us not sweep the latest builds for all packages, which will help us in scenarios like we faced today (april 5th) when preparing 4.6 advisories. Which was that we had nightlies for 4.6 already that we wanted to use. But before automation could be frozen, ocp4 job triggered and started building again, it build some packages and then failed for some. Brew outage also responsible for other failures. 

The point is even if we had all success builds, we would still want the option of using builds from a point of time in the past. And in cases like partial builds succeeding, this would be critical.

---

In order to sweep latest builds for a (tag, package) combo we need to give an event id (similar to a timestamp) to brew methods. Luckily the method we are using for sweeping image builds supports it. 

This event id can be found by multiple ways, one of which is the creation_event_id of getBuild()

Todo:
[X] works with -k image
[X] works with -k rpm

Test
- `brew list-history --tag=rhaos-4.6-rhel-8-candidate --after="2021-04-27" -e`
- `./elliott -g openshift-4.6 find-builds -k image --brew-event 38753378`
- `./elliott -g openshift-4.6 find-builds -k rpm --brew-event 38753378`

### DIFF
--- a/elliottlib/cli/find_builds_cli.py
+++ b/elliottlib/cli/find_builds_cli.py
@@ -305,7 +305,7 @@ def _fetch_builds_by_kind_rpm(tag_pv_map, brew_event, brew_session):
         else:
             red_print("key of brew_tag_product_version_mapping in erratatool.yml must be candidate\n")
             continue
-        candidates = elliottlib.brew.find_unshipped_build_candidates(base_tag, brew_event, brew_session)
+        candidates = elliottlib.brew.find_unshipped_build_candidates(base_tag, brew_event, kind='rpm', brew_session=brew_session)
         rpm_tuple.extend(_gen_nvrp_tuples(candidates, tag_pv_map, tag))
     return rpm_tuple
 
--- a/elliottlib/cli/find_builds_cli.py
+++ b/elliottlib/cli/find_builds_cli.py
@@ -70,7 +70,7 @@ pass_runtime = click.make_pass_decorator(Runtime)
     help='Only attach non-payload images')
 @click.option(
     '--brew-event', required=False,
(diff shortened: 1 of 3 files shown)


thegreyd:
Most of my time was spent (see: wasted) trying to find about the event id and brew source code :facepalm: 
Took me an embarasingly long time to realize python koji was a thin client. 
Here's the api - https://koji.fedoraproject.org/koji/api
Here's the source code for the server https://pagure.io/koji/blob/master/f/hub/kojihub.py where you can ctrl-f the methods

sosiouxme:
`flake8` error there, also wanted to note you can get an event id from `brew list-history -v`.

elliottlib/cli/find_builds_cli.py
```diff
@@ -255,7 +260,7 @@ def _fetch_builds_from_diff(from_payload, to_payload, tag_pv_map):
     return _fetch_nvrps_by_nvr_or_id(nvrs, tag_pv_map)
 
 
-def _fetch_builds_by_kind_image(runtime, tag_pv_map, brew_session, p, np):
+def _fetch_builds_by_kind_image(runtime, tag_pv_map, brew_event, brew_session, p, np):
```
sosiouxme, line 263: we had it for RPMs already but not images? bizarre.

sosiouxme (approved):
alright looks solid to me!

next stop, bugs :)

/lgtm

elliottlib/cli/find_builds_cli.py
thegreyd: Added it to rpm now

PR #188, opened 2021-04-19
ART-2718: When switching to RHSA set the appropriate Impact in the topic

jasinner:
Requires changes to ocp-build-data to add ${wildcards} to erratatool.yaml. It won't break with existing ocp-build-data though, just won't actually replace the correct impact in topic.


sosiouxme:
Thanks a lot for taking the initiative on this.

The complication with ART-2718 is that we already have a [very simple template system in place](https://github.com/openshift/elliott/blob/98bc884b06c194d2dace9af70f9e7e5f749168c4/elliottlib/gitdata.py#L220-L226) when reading these files ([see for example](https://github.com/openshift/ocp-build-data/blob/openshift-4.8/erratatool.yml#L9)), and it happens to already use curly braces (really it's just python format function), and it doesn't like when it can't find an expected value, so it would choke on `${impact}`. We should probably switch to a real template system, but that's a larger project.

So, we can either augment the vars being fed into reading that file to calculate and include `impact` (always, even if no RHSAs), or we can have a second lame "template" system that just does something like `topic.replace('[[impact]]', impact)`, or maybe we can think of something cleaner.

jasinner:
Pushed another commit which attempts to switch the existing template function to use Mako as well.
I think we'll need a proper templating feature in order to do https://issues.redhat.com/browse/ART-2720

elliottlib/template.py
```diff
@@ -0,0 +1,20 @@
+from __future__ import absolute_import, print_function, unicode_literals
+from mako.template import Template
+
+
+def render_topic(template, impact="Important"):
+    t = Template(template)
+    return t.render(impact=impact)
+
+
+def render(template, MAJOR, MINOR):
```
vfreex, line 10: I think this function should take `kwargs` rather than hardcoded `MAJOR` `MINOR` arguments.

vfreex (changes requested):
Generally LGTM except some nits.

vfreex:
@jasinner Could you update your PR?
The issue is that variables are defined in group.yml or Elliott/Doozer internally and we will have more variables. `render` sounds more generic than just replacing `MAJOR` and `MINOR`. Once this is addressed we can merge, test, and improve this.

elliottlib/template.py
jasinner: Fixed, please review again.

jasinner:
> @jasinner Could you update your PR?
> The issue is that variables are defined in group.yml or Elliott/Doozer internally and we will have more variables. `render` sounds more generic than just replacing `MAJOR` and `MINOR`. Once this is addressed we can merge, test, and improve this.

done.

jasinner:
I'm going to close this PR, as we need a better way to add templating which doesn't conflict with the existing template system used by Dozer and CI.

As mentioned by ART, we could probably use the existing format system ({} without $) and pre-fill the vargs in group-vars.

PR #189, opened 2021-05-21
ART-2927: find-builds basic assembly awareness

vfreex:
- If assemblies feature is disabled, find-builds should keep the current
  behavior (attaching latest builds in -candidate Brew tags).

- Running elliott, without an assembly parameter should only attach images/rpms to advisories if they have .assembly.stream or, if no .assembly.stream exists, then the latest build without .assembly in the RELEASE field.

- Running elliott with an assembly parameter should prefer builds with that assembly in the RELEASE field.

### DIFF
--- a/.devcontainer/dev.Dockerfile
+++ b/.devcontainer/dev.Dockerfile
@@ -1,4 +1,4 @@
-FROM fedora:33
+FROM fedora:34
 LABEL name="elliott-dev" \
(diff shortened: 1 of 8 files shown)


vfreex:
/hold

Hold this until figuring out which brew tags needs to be queried when `runtime_assembly == "stream"` and `runtime_assembly != "stream"`

PR #191, opened 2021-06-08
ART-2718: When switching to RHSA set the appropriate Impact in the topic

jasinner:
Reworking of https://github.com/openshift/elliott/pull/188 to use existing templating feature of ocp_build_data.

### DIFF
--- a/elliottlib/cli/attach_cve_flaws_cli.py
+++ b/elliottlib/cli/attach_cve_flaws_cli.py
@@ -86,16 +86,17 @@ def attach_cve_flaws_cli(runtime, advisory_id, noop, default_advisory_type):
         runtime.logger.info('No "first-fix" bugs found, exiting')
         exit(0)
 
+    cve_boilerplate = runtime.gitdata.load_data(key='erratatool').data['boilerplates']['cve']
+
     advisory = Erratum(errata_id=advisory_id)
     if not is_security_advisory(advisory):
         runtime.logger.info('Advisory type is {}, converting it to RHSA'.format(advisory.errata_type))
-        cve_boilerplate = runtime.gitdata.load_data(key='erratatool').data['boilerplates']['cve']
         advisory.update(
             errata_type='RHSA',
             security_reviewer=cve_boilerplate['security_reviewer'],
             synopsis=cve_boilerplate['synopsis'],
             description=cve_boilerplate['description'],
-            topic=cve_boilerplate['topic'],
+            topic=cve_boilerplate['topic'].format(IMPACT="Low"),
             solution=cve_boilerplate['solution'],
             security_impact='Low',
         )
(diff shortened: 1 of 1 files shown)


elliottlib/cli/attach_cve_flaws_cli.py
```diff
@@ -115,6 +116,11 @@ def attach_cve_flaws_cli(runtime, advisory_id, noop, default_advisory_type):
         ))
         advisory.update(security_impact=highest_impact)
 
+    if highest_impact not in advisory.topic:
```
vfreex, line 119: Line 117 only updates the security impact when the current impact of advisory is smaller than `highest_impact`. I think the topic should be updated together with advisory security impact change. Any thought?

vfreex (changes requested):
Thanks for the PR! Generally LGTM, only 1 question. See my comment inline.

vfreex:
Only flake8 needs to be fixed.

elliottlib/cli/attach_cve_flaws_cli.py
jasinner: Agreed, it's better to always set to the highest_impact. I removed the if statement here.

jasinner:
> Only flake8 needs to be fixed.

Fixed now. Thanks.

elliottlib/cli/attach_cve_flaws_cli.py
```diff
@@ -128,7 +133,7 @@ def attach_cve_flaws_cli(runtime, advisory_id, noop, default_advisory_type):
 
     if noop:
         print('DRY-RUN: The following changes would have been applied to the advisory:')
-        print(advisory)
+        print(advisory.dump())
```
sosiouxme, line 136: i observed `advisory.dump()` seems to have a bug:
```
  File "/workspaces/elliott/elliottlib/cli/attach_cve_flaws_cli.py", line 136, in attach_cve_flaws_cli
    print(advisory.dump())
  File "/home/lmeyer/.local/lib/python3.9/site-packages/errata_tool/erratum.py", line 996, in dump
    print("Manager Email: " + self.manager_email)
TypeError: can only concatenate str (not "NoneType") to str
```
`manager_email` should be set there. that's nothing wrong with this code, just something we might want to consider reverting if it causes problems.

sosiouxme (approved):
LGTM :ship: 

and thanks for the help!

elliottlib/cli/attach_cve_flaws_cli.py
jasinner: Ah yes, I remember hitting from problem now. I think I removed offending line from erratum.py but that means it doesn't print the manager_email. If we don't use dump we don't get the updated description section of the erratum.

PR #193, opened 2021-06-14
find-builds: Fix exception in log output

joepvd:
```
...
    click.echo(f'Found {len(brew_latest_builds)} builds. Filtering out shipped builds...')
TypeError: object of type 'generator' has no len()
```


thiagoalessio (approved):
/lgtm

PR #194, opened 2021-06-14
[Sweep] Attach bugs in batches

thegreyd:
When we prepare GA advisories, we attach bugs in numbers exceeding a thousand. Errata tool attaches bugs but fails on the part where it posts a single comment about all the bugs attached in a request. The single comment is huge and it exceeds the character limit defined. This is a change so that doesn't happen.


joepvd (approved):
/lgtm

thiagoalessio:
ooh we merged it too early. flake8 validation failed, which prevented the tests to run.

PR #195, opened 2021-06-14
Fix find-builds -k rpm

joepvd:
Slurping a generator for a second time makes one inhale nothingness rather than
consume content. With this change, rpms will be attached to the advisory.


thiagoalessio (approved):
/lgtm

PR #199, opened 2021-06-21
ART-1155 Waive multiple rpmdiff failures

thegreyd:
https://issues.redhat.com/browse/ART-1155

./elliott -g openshift-4.8 rpmdiff show 77463 --waive "Waiving for now" --dry-run


vfreex:
This is actually something I wanted to do for a long time but the team worries about the risk of waiving things using such a command without thinking...

PR #203, opened 2021-06-30
WIP: ART-3099 Sweep ignores placeholder bugs

thegreyd:
Sometimes when an advisory is dropped because of no shipping bugs/rpms/images, the placeholder bug lingers on if it is not closed, and unintentionally attached to an advisory. So let's avoid placeholder bugs in sweep

Another need when preparing GA advisories is to get accurate sweep dry-run output. Right now dry-run doesn't account for bugs that will fail to be attached that are already attached to another advisory. So we try to fetch advisory link from proposed bugs and filter out those that will fail anyways to provide a more accurate dry-run output.

### DIFF
--- a/elliottlib/errata.py
+++ b/elliottlib/errata.py
@@ -414,50 +414,69 @@ def parse_exception_error_message(e):
     return [int(b.split('#')[1]) for b in re.findall(r'Bug #[0-9]*', str(e))]
 
 
-def add_bugs_with_retry(advisory, bugs, noop=False, batch_size=100):
+def fetch_advisory(advisory_id):
+    try:
+        advisory = Erratum(errata_id=advisory_id)
+    except GSSError:
+        exit_unauthenticated()
+
+    if advisory is False:
+        raise exceptions.ElliottFatalError(f"Error: Could not locate advisory {advisory_id}")
+    return advisory
+
+
+def filter_advisory_bugs(advisory, bugs, bzapi):
+    # filter existing advisory bugs
+    existing_bugs = advisory.errata_bugs
+    new_bugs = set(bug.id for bug in bugs) - set(existing_bugs)
+
+    # filter placeholder bugs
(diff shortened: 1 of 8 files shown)


elliottlib/errata.py
```diff
@@ -414,7 +414,7 @@ def parse_exception_error_message(e):
     return [int(b.split('#')[1]) for b in re.findall(r'Bug #[0-9]*', str(e))]
 
 
-def add_bugs_with_retry(advisory, bugs, noop=False, batch_size=100):
+def add_bugs_with_retry(advisory, bugs, bzapi, noop=False, batch_size=100):
```
vfreex, line 417: I would prefer to filter bugs that are already attached to an advisory out before calling `add_bugs_with_retry` rather than filter in `add_bugs_with_retry`. That will decouple the BZ api calls from errata module and keep `add_bugs_with_retry` simpler. Thoughts?

vfreex (changes requested):

elliottlib/errata.py
vfreex: Another reason to do the filtering out of this function is that `add_bugs_with_retry` should sound what it is like. If it fails to attach specified bugs to specific advisory, it should raise an exception so we know there's an error. If we have the filter in this function, it will just return without any error even if the already attached bugs are on a different advisory, which looks like an unexpected behavior to me.

vfreex (changes requested):

vfreex:
To sum up, I would prefer to pass the correct arguments to a function rather than pass the incorrect ones then silently ignore the error inside the function.

elliottlib/errata.py
thegreyd: +1 I have refactored and moved this to another function

vfreex:
@thegreyd could you rebase this PR?

PR #204, opened 2021-07-01
ART-3045: Assembly lense

vfreex:
- Copies assembly.py from https://github.com/openshift/doozer/pull/431.
- Loads group config from releases.yml
- Loads basis event from releases.yml

Note:
1. Computing cutoff timestamp from basis event in `find-bugs` is not covered.
2. Honoring `is` in `find-builds` is not covered.
3. I didn't introduce the global `--brew-event` option because Elliott doesn't have a global koji wrapper that injecting global event Id like Doozer. If we need that, we can introduce that in a separate PR.
4. Please also look at https://github.com/openshift/ocp-build-data/pull/979. It adds add -hotfix tag to Errata product version mapping. We need that mapping to attach builds.

### DIFF
--- a/elliottlib/assembly.py
+++ b/elliottlib/assembly.py
@@ -1,0 +1,173 @@
+import copy
+import typing
+
+from elliottlib.model import Missing, Model
+
+
+def merger(a, b):
+    """
+    Merges two, potentially deep, objects into a new one and returns the result.
+    Conceptually, 'a' is layered over 'b' and is dominant when
+    necessary. The output is 'c'.
+    1. if 'a' specifies a primitive value, regardless of depth, 'c' will contain that value.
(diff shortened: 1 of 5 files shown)


jupierce (approved):
Should we consider a single repo `art-tools` for doozer and elliott? Just a natural discomfort with duplicating code.

vfreex:
@jupierce I think having a common shared lib for both elliott and doozer (and maybe other tools) in the future is likely to be necessary. Also if you do a line by line comparison for this PR to the Doozer code, there's some difference because some shared code in elliott and doozer are not in sync and sometimes the same class may have slightly different methods and fields. 
Would like to have that in the future, but I guess both elliott and doozer will require some code refactoring.

PR #210, opened 2021-07-07
Add validate-rhsa cli command

sfowl:
Example usage:

```
$ elliott validate-rhsa 77087
http://sfm2.prodsec.redhat.com/api/public/errata/77087/alerts
Validating..

E: Erratum does not contain link to Release/Technical Notes in References
  > If an erratum refers to Technical or Release Notes in the Problem
  > Description, an accompanying URL must be present in the References section.

$ elliott validate-rhsa 99999
http://sfm2.prodsec.redhat.com/api/public/errata/99999/alerts
Validating..

HTTP 500: Could not validate advisory 99999 (is it an RHSA? without embargo?)
```
In the above, 77087 is an advisory ID that has already shipped and 99999 is an invalid advisory ID. This command will be most useful for RHSAs that have not shipped yet. It prints the Alerts that ProdSec also sees when they are reviewing.


elliottlib/cli/validate_advisory.py
```diff
@@ -0,0 +1,44 @@
...
+    session = requests.Session()
+    url = constants.SFM2_ERRATA_ALERTS_URL.format(id=advisory)
+    print(url)
+    print("Validating..")
+    print()
+    resp = session.get(url)
+
+    # Will fail if id invalid, not an RHSA or RHSA has embargoed data
+    if resp.status_code != 200:
+        print(f"HTTP {resp.status_code}: Could not validate advisory {advisory} (is it an RHSA? without embargo?)")
```
joepvd, line 32: Is there a period of time where the server is acquiring the data, and would a retry make sense here?
Would it make sense to query erratatool to disambiguate between the two (maybe three) reasons for status to be != 200?

joepvd (commented):
Nice!

elliottlib/cli/validate_advisory.py
sfowl: When the server receives a request, it will reload the advisory from ET before responding. So the returned Alerts should always be up-to-date.

I guess we could add an additional query to ET on error, but I'm not sure if that additional query/response time is worth the dis-ambiguity of errors.

elliottlib/cli/validate_rhsa.py
```diff
@@ -0,0 +1,44 @@
...
+    url = constants.SFM2_ERRATA_ALERTS_URL.format(id=advisory)
+    print(url)
+    print("Validating..")
+    print()
+    resp = session.get(url)
+
+    # Will fail if id invalid, not an RHSA or RHSA has embargoed data
+    if resp.status_code != 200:
+        print(f"HTTP {resp.status_code}: Could not validate rhsa {advisory} (is it an RHSA? without embargo?)")
+        return
```
vfreex, line 33:
```suggestion
        exit(1)
        return
```

elliottlib/cli/validate_rhsa.py
```diff
@@ -0,0 +1,44 @@
...
+    alerts = resp.json()
+
+    if alerts:
+        for a in alerts:
+            print("E: {text}".format(text=a.get("text")))
+            howto = a.get("how_to_resolve")
+            if howto:
+                print(textwrap.indent(howto, "  > "))
+            print()
```
vfreex, line 42:
```suggestion
            print()
            exit(2)
```

vfreex (commented):
Generally LGTM. Would like to add exit code so we can integrate this command in our automation in the future.

vfreex (approved):
/lgtm

PR #211, opened 2021-07-08
ART-3095 Flaw bugs with no additional trackers are considered first fixes

thegreyd:
In [my last PR](https://github.com/openshift/elliott/pull/201) we fixed the issue of not failing 
but the function return value wasn't accurate.
If additional trackers are not found for the flaw bug
then the flaw bug IS considered a first fix.

[relevant slack thread](https://coreos.slack.com/archives/CB95J6R4N/p1625775501275100)

### DIFF
--- a/elliottlib/attach_cve_flaws.py
+++ b/elliottlib/attach_cve_flaws.py
@@ -34,24 +34,27 @@ def get_corresponding_flaw_bugs(bzapi, tracker_bugs):
     return [flaw_bug for flaw_bug in blocking_bugs if bzutil.is_flaw_bug(flaw_bug)]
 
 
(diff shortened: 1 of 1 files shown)


sfowl:
@thegreyd The first fix check needs to checking if it is a first fix for the given *componnent* (i.e. rpm or container). I vaguely recall another item for this in the ART backlog, but not sure it was completed yet.

vfreex:
@sfowl I vaguely remember the component name check is already done when elliott attaches trackers. So this LGTM but I could be wrong.

vfreex (approved):

thegreyd:
Thanks Sam, here's the story https://issues.redhat.com/browse/ART-2725
Since this PR is fixing a regression (that went in with an earlier PR in the description), I'd like to merge this,
and get the other story in with another PR.

PR #212, opened 2021-07-12
ART-2725 Add component checks to first fix

thegreyd:
https://issues.redhat.com/browse/ART-2725
[relevant thread](https://coreos.slack.com/archives/CB95J6R4N/p1625775501275100)

Earlier we were filtering out flaw bugs
based on if any of its tracker bugs were 
fixed/shipped earlier. Now we determine
it based on components. And if for any 
component, a tracker was not fixed/shipped
then we do not filter out the flaw bug.

And add missing test cases, so we know when we regress!


sfowl:
lgtm

PR #216, opened 2021-07-21
default advisories with list mode

thegreyd:
When preparing releases, sometimes we have new nightlies, when new bugs/builds need to be attached to advisories. In this case we don't want to use sweep, but still be able to provide a list of bugs to be attached. For that we have list mode.

In sweep mode we have filtering to determine which bugs go where (rpm component, extras components), which is not present in list. This change brings that to list mode

`elliott -g openshift-4.8 find-bugs --mode list --id 1981871,1970975,1981633,1982250,1982252,1980788,1981871 --into-default-advisories --dry-run`


thiagoalessio (approved):
/lgtm

vfreex (approved):
/lgtm

PR #217, opened 2021-07-22
[ART-3175] Add "Automation" keyword to placeholder bugs

thiagoalessio:
So they can be excluded from reports.


thiagoalessio:
https://github.com/python-bugzilla/python-bugzilla/blob/370eddbe175bb24a4abb37d2fd45fe75219b6c44/bugzilla/base.py#L1722

sosiouxme:
Where's the test case? :-P

PR #224, opened 2021-08-26
art-3078: Add create text only cli

Ximinhan:
Add create text only cli for elliott it create a custom advisory and along with a custom bug in create text only advisory use case, all required data for advisory and bug need pass from jenkins job's parameters since each cases are different.

### DIFF
--- a/elliottlib/cli/create_textonly_cli.py
+++ b/elliottlib/cli/create_textonly_cli.py
@@ -1,4 +1,3 @@
-from __future__ import absolute_import, print_function, unicode_literals
 import datetime
 
 from errata_tool import Erratum, ErrataException
@@ -100,12 +99,9 @@ def create_textonly_cli(ctx, runtime, errata_type, date, assigned_to, manager, p
             bug_id=newbug.id,
             text_only=1,
         )
-    except elliottlib.exceptions.ErrataToolUnauthorizedException:
-        exit_unauthorized()
     except elliottlib.exceptions.ErrataToolError as ex:
         raise ElliottFatalError(getattr(ex, 'message', repr(ex)))
 
-    # erratum.addBugs(newbug.id)
     if yes:
         erratum.commit()
         green_prefix("Created new text only advisory: ")
--- a/elliottlib/cli/create_textonly_cli.py
+++ b/elliottlib/cli/create_textonly_cli.py
@@ -99,8 +99,10 @@ def create_textonly_cli(ctx, runtime, errata_type, date, assigned_to, manager, p
             bug_id=newbug.id,
             text_only=1,
         )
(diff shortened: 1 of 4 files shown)


elliottlib/cli/create_textonly_cli.py
```diff
@@ -0,0 +1,116 @@
...
+            manager_email=manager,
+            date=release_date,
+            bug_id=newbug.id,
+            text_only=1,
+        )
+    except elliottlib.exceptions.ErrataToolUnauthorizedException:
+        exit_unauthorized()
+    except elliottlib.exceptions.ErrataToolError as ex:
+        raise ElliottFatalError(getattr(ex, 'message', repr(ex)))
+
```
vfreex, line 107: No need to capture this exception and re-raise again.

elliottlib/cli/create_textonly_cli.py
```diff
@@ -0,0 +1,116 @@
...
+    # erratum.addBugs(newbug.id)
```
vfreex, line 108: Commented code needs to be removed.

elliottlib/cli/create_textonly_cli.py
```diff
@@ -0,0 +1,116 @@
...
+
+pass_runtime = click.make_pass_decorator(Runtime)
+
+#
+# Create Text only advisory
+# advisory:create-textonly
+
+
+@cli.command('create-textonly',
+             short_help='Create a textonly advisory along with notification bug attached')
```
vfreex, line 25: I am OK with a new command, but would prefer to integrate this into `create` command with a new impetus so that we can leverage the advisory boilerplate text in group.yml.

elliottlib/cli/create_textonly_cli.py
```diff
@@ -0,0 +1,116 @@
+from __future__ import absolute_import, print_function, unicode_literals
```
vfreex, line 1: Python 3 doesn't need this.

vfreex (changes requested):
Basically LGTM. Some nits.

elliottlib/cli/create_textonly_cli.py
Ximinhan: I used to try to use a template in erratatool.yml but I realized in this use case for text-only we need to update almost every comment for each release since it's has different reasons, it's hardly create with boilerplate without change. Also the same reason to the related bug, so I thought the default value for jenkins job's parameters can treat as a boilerplate.

Ximinhan: For each text-only advisory we need to update the context about effected version, target version, related blocker bug and related blog/source link.

PR #225, opened 2021-08-27
[Hack n Hustle] Get and update common fields across advisories

thegreyd:
I wanted a command to update the shipping date of a group of advisories without having to go through ET.

Made it a little general purpose so you can get any field across advisories, if you wanted to.
And then update common fields across advisories like date, manager, qe, owner etc. 
I don't see us updating the others as much
and date as basically the only real feature :) 
but worth it

### DIFF
--- a/elliottlib/cli/advisory_commons_cli.py
+++ b/elliottlib/cli/advisory_commons_cli.py
@@ -1,0 +1,156 @@
+from elliottlib import logutil
+from elliottlib.cli.common import cli
+from elliottlib.util import green_prefix
+from errata_tool import Erratum, ErrataException
+import click
+import string
+import re
+
+LOGGER = logutil.getLogger(__name__)
+supported_update_fields = [
+    'publish_date',
+    'text_only',
+    'text_only_cpe',
+    'owner_email',
+    'manager_email',
+    'manager_id',
+    'qe_email',
+    'qe_group',
+]
+
+
+@cli.command("advisory-commons", short_help="Show or update advisory commons for a group of advisories")
+@click.argument("advisories", nargs=-1, type=click.IntRange(1), required=False)
+@click.option("--field",
+              help="Advisory field (like publish_date) to show or update. Examples of show only fields: ["
+                   "'errata_state', "
+                   "'errata_name', 'content_types', 'synopsis', "
+                   "'errata_bugs', "
+                   "'errata_type']")
+@click.option("--new",
+              help="New value of field to update advisories with. --field should be one of"
+                   f" {supported_update_fields}")
+@click.option("--version-replace",
+              help="<existing_version>:<new_version>"
+                   "Update synopsis, topic, description fields of advisories with new version."
+                   "Example --version-replace '4.7.26:4.7.28' ")
+@click.option('--yes', '-y', is_flag=True,
+              default=False, type=bool,
+              help="Update the advisories (by default only a preview is displayed)")
+@click.pass_obj
+def advisory_commons_cli(runtime, advisories, field, new, version_replace, yes):
+    """Display or Change a common field (like date) across multiple advisories.
+
+Advisories created for an OCP version have common fields, that sometimes
+will need updating. This command helps with that.
+
+    NOTE: The two advisory input options (--assembly and --advisories)
+    are mutually exclusive and can not be used together.
+
+    Show the field "publish_date" for all advisories for an assembly/group/advisories
+
+    $ elliott -g openshift-4.8 --assembly 4.8.8 advisory-commons --field "publish_date"
+
+    $ elliott -g openshift-3.11 advisory-commons --field "publish_date"
+
+    $ elliott advisory-commons 80825 80824 --field "publish_date"
+
+    (Preview) update field "publish_date" for all advisories for an assembly
+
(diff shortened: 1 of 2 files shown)


elliottlib/cli/advisory_commons_cli.py
```diff
@@ -0,0 +1,103 @@
+from __future__ import absolute_import, print_function, unicode_literals
```
vfreex, line 1: Python 3 doesn't need this.

elliottlib/cli/advisory_commons_cli.py
```diff
@@ -0,0 +1,103 @@
...
+              help="Select given advisory id(s). comma separated")
+@click.option("--field", required=True,
+              help="Advisory field (like publish_date) to show or update. Examples of show only fields: ["
+                   "'errata_state', "
+                   "'errata_name', 'content_types', 'synopsis', "
+                   "'errata_bugs', "
+                   "'errata_type']")
+@click.option("--new",
+              help=f"New value of field to update advisories with. --field should be one of"
+                   f" {supported_update_fields}")
```
vfreex, line 32: Would like to have separate options for each supported field for better validation, help message, and being able to update multiple fields.

elliottlib/cli/advisory_commons_cli.py
```diff
@@ -0,0 +1,103 @@
...
+    'manager_email',
+    'manager_id',
+    'qe_email',
+    'qe_group',
+]
+
+
+@cli.command("advisory-commons", short_help="Show or update advisory commons for a group of advisories")
+@click.option("--advisories", metavar='ADVISORY',
+              help="Select given advisory id(s). comma separated")
```
vfreex, line 23: Guess passing advisories as arguments in stead of options would be more ergonomics?
```suggestion
@click.argument("advisories", nargs=-1, type=click.IntRange(1), required=False)
```

elliottlib/cli/advisory_commons_cli.py
```diff
@@ -0,0 +1,103 @@
...
+                    green_prefix(f"No Change: {advisory_id}: ")
+                    click.echo(f"Target field state is same as current state: {field}: {current}")
+                else:
+                    if noop:
+                        green_prefix(f"NOOP: {advisory_id}: ")
+                        click.echo(f"Would have changed field {field} {current} ➔ {new}")
+                    else:
+                        update_kwarg = {}
+                        if field == 'publish_date':
+                            update_kwarg['date'] = new
```
vfreex, line 89: I didn't see a compelling reason to rename this field. We have called this field `date` or `release_date` elsewhere in Elliott code.

elliottlib/cli/advisory_commons_cli.py
```diff
@@ -0,0 +1,103 @@
...
+
+    $ elliott -g openshift-4.8 --assembly 4.8.8 advisory-commons --field "publish_date" --new "2021-Aug-31"
+
+    Do not actually change anything
+
+    $ elliott -g openshift-4.8 --assembly 4.8.8 advisory-commons --field "publish_date" --new "2021-Aug-31" --noop
+"""
+    count_flags = sum(map(bool, [runtime.assembly != 'stream', advisories]))
+    if count_flags > 1:
+        raise click.BadParameter("Use only one of --assembly or --advisories")
```
vfreex, line 61: For releases without assembly support (like 3.11), `runtime.assembly` is `None`.

BTW I don't think this check is absolutely needed.

vfreex (commented):
Thank you for adding this useful command in Hack'n'Hustle. Just some nits.

joepvd:
Nice! Would be great with a search-and-replace for the title in case, say, 4.8.8 gets renamed as 4.8.9.

elliottlib/cli/advisory_commons_cli.py
thegreyd: The weird thing is errata_tool's advisory.update() method searches for "date" specifically, any other field name is ignored.

elliottlib/cli/advisory_commons_cli.py
thegreyd: Yes! I didn't know argument could be required=False :) that is definitely better.

elliottlib/cli/advisory_commons_cli.py
thegreyd: Maybe next hack n hustle :sweat_smile: ?
Don't want to spend too much time on this right now.. date format is validated in errata_tool which I think is sufficient for the initial implementation

elliottlib/cli/advisory_commons_cli.py
thegreyd: fixed!

thegreyd:
> Nice! Would be great with a search-and-replace for the title in case, say, 4.8.8 gets renamed as 4.8.9.
> 
> Apologies for the scope creep ;)

I added a --version-replace option as a shortcut to bump versions. PTAL
Noticed that we can't do a blanket search and replace in description (because of bug titles being part of it), so I added template lines

elliottlib/cli/advisory_commons_cli.py
thegreyd: Added a --version-replace option specifically for topic, synopsis and description

vfreex:
LGTM once you fix the conflict.

joepvd (approved):
/lgtm

Maybe squash or rebase into a single commit.

PR #228, opened 2021-09-10
Fix a find-builds issue with assemblies

vfreex:
This issue occurs when we attach kernel rpms defined as group
dependencies to 4.8.11 advisory.

Unlike member rpms, group dependencies are packages that are not built or
managed by ART. We need to check if they have valid `rhaos-` tags and
attach them if they do.


sosiouxme:
LGTM too

PR #230, opened 2021-09-16
Bugzilla: Query again if results equals limit

joepvd:
Until https://github.com/python-bugzilla/python-bugzilla/issues/149 is
resolved and the python-bugzilla library takes care of pagination,
implement limit/offset searching here.


sosiouxme:
@joepvd flake8 doesn't like it

joepvd:
Now at least I am seeing the same error locally.

thiagoalessio (approved):
/lgtm

PR #231, opened 2021-09-30
ART-2720 Automate description update at RHSA conversion

thegreyd:
https://issues.redhat.com/browse/ART-2720

This will require update to erratatool.yml in all current branches. So don't merge before all those are updated.


sosiouxme (approved):
/lgtm

i don't think this would break anything without an update in `erratatool.yml` - the description content there just wouldn't be dynamic.

jasinner:
/lgtm

Thanks

thiagoalessio (approved):
/lgtm

PR #232, opened 2021-10-01
Deprecate create rhsa 

thegreyd:
This came out of a discussion here https://github.com/openshift/ocp-build-data/pull/1104#discussion_r719812266 re rhsa process update https://github.com/openshift/elliott/pull/231

Our rhsa advisory process flow has moved from
directly creating rhsa -> creating rhba and then updating it to rhsa
Since we are no longer doing that it seems appropriate to retire the functionality


joepvd (approved):
/lgtm

PR #233, opened 2021-10-05
Fail if placeholder cannot get substituted

joepvd:
In case there is a substitution where no key is provided, leave it alone


sosiouxme:
[doozer just lets it fail](https://github.com/openshift/doozer/blob/61132a318cf7e48391db648c8dc6aa675015f71f/doozerlib/runtime.py#L301-L302) and that seems like the right approach to me.

PR #234, opened 2021-10-12
Fix CVE tracker bug sweep

joepvd:
A tracker bug for component with a `.` in its name could not be swept
into the advisory, as it could not be verified that a build with that
name was attached. This commit relaxes the regex, and includes also
`dots` in the component name.


thiagoalessio (approved):
/lgtm

PR #240, opened 2021-10-27
[ART-3477] Fail attach-cve if tracker has no flaw

joepvd:


elliottlib/attach_cve_flaws.py
```diff
@@ -52,7 +52,19 @@ def get_corresponding_flaw_bugs(bzapi, tracker_bugs, fields):
...
+    flaw_bugs = [flaw_bug for flaw_bug in blocking_bugs if bzutil.is_flaw_bug(flaw_bug)]
+
+    # Validate that each tracker has a corresponding flaw bug
+    flaw_ids = set([b.id for b in flaw_bugs])
+    no_flaws = set()
+    for tracker in tracker_bugs:
+        if len(set(tracker.blocks).intersection(flaw_ids)) == 0:
+            no_flaws.add(tracker.id)
+    if len(no_flaws) > 0:
+        raise exceptions.ElliottFatalError(f'No flaw bugs could be found for these trackers: {no_flaws}')
```
vfreex, line 65: Would like to move this check out of this function. I think I used this function elsewhere so don't really want to break things.

vfreex (changes requested):

elliottlib/attach_cve_flaws.py
```diff
@@ -52,7 +54,22 @@ def get_corresponding_flaw_bugs(bzapi, tracker_bugs, fields):
...
+
+    flaw_bugs = [flaw_bug for flaw_bug in blocking_bugs if bzutil.is_flaw_bug(flaw_bug)]
+
+    # Validate that each tracker has a corresponding flaw bug
+    flaw_ids = set([b.id for b in flaw_bugs])
+    no_flaws = set()
+    for tracker in tracker_bugs:
+        if len(set(tracker.blocks).intersection(flaw_ids)) == 0:
+            no_flaws.add(tracker.id)
+    if len(no_flaws) > 0:
```
sosiouxme, line 66: we have it [on authority from pep 8](https://www.quora.com/What-is-the-Pythonic-way-to-check-for-an-empty-set) not to use `len` for this :)
```suggestion
        if not set(tracker.blocks).intersection(flaw_ids):
            no_flaws.add(tracker.id)
    if no_flaws:
```

sosiouxme (commented):
Otherwise LGTM, thanks!

vfreex (approved):

vfreex:
PR needs rebase

PR #241, opened 2021-11-02
[ART-2880] enhance repair-bugs and remove-bugs

Ximinhan:
add auto option for remove-bugs
add comment option and close-placeholder option for repair-bugs


elliottlib/cli/__main__.py
```diff
@@ -113,18 +117,17 @@ def remove_bugs(runtime, advisory, default_advisory_type, id):
 
 
 """
+    if auto and len(id) > 0:
+        raise click.BadParameter("Combining the automatic and manual bug modification options is not supported")
+    if not auto and len(id) == 0:
```
sosiouxme, line 122: i prefer the [pep 8 style of checking for empty](https://www.quora.com/What-is-the-Pythonic-way-to-check-for-an-empty-set):
```suggestion
    if auto and id:
        raise click.BadParameter("Combining the automatic and manual bug modification options is not supported")
    if not auto and not id:
```

elliottlib/cli/__main__.py
```diff
@@ -96,8 +96,12 @@
 @click.option("--id", metavar='BUGID', default=[],
               multiple=True, required=True,
```
sosiouxme, line 97: can no longer be required
```suggestion
              multiple=True, required=False,
```

elliottlib/cli/__main__.py
```diff
@@ -138,6 +141,12 @@ def remove_bugs(runtime, advisory, default_advisory_type, id):
             raise ElliottFatalError("Error: Could not locate advisory {advs}".format(advs=advisory))
 
         try:
+            if auto:
+                bug_ids = advs.errate_bugs
```
sosiouxme, line 145:
```suggestion
                bug_ids = advs.errata_bugs
```

elliottlib/cli/__main__.py
```diff
@@ -138,6 +141,12 @@ def remove_bugs(runtime, advisory, default_advisory_type, id):
...
+            else:
+                bug_ids = [bzapi.getbug(i) for i in cli_opts.id_convert(id)]
```
sosiouxme, line 147: it looks like your goal here is to get not bug ids, but actual bug records. so i would call them just `bugs`
```suggestion
                bugs = advs.errata_bugs
            else:
                bugs = [bzapi.getbug(i) for i in cli_opts.id_convert(id)]
```
... although the error i get when playing with this indicates to me that `advs.errata_bugs` is probably a list of IDs.

```
$ ./elliott --group openshift-4.9 remove-bugs -a 83078 --auto
[...]
  File "/workspaces/elliott/elliottlib/cli/__main__.py", line 149, in <listcomp>
    click.echo(" {}".format(", ".join([str(b.bug_id) for b in bug_ids])))
AttributeError: 'int' object has no attribute 'bug_id'
```

Also, it's not likely to be important here, but if there is a way to retrieve many bugs at once from the bzapi, rather than one at a time... that would be nice.

elliottlib/cli/__main__.py
```diff
@@ -259,13 +268,20 @@ def get(ctx, runtime, default_advisory_type, details, id_only, as_json, advisory
               default='ON_QA',
               type=click.Choice(elliottlib.constants.VALID_BUG_STATES),
               help="Final state of the bugs (default: ON_QA)")
+@click.option("--comment", "comment",
+              required=False,
+              help="Add comment to bug")
+@click.option("--colse-placeholder", "close_placeholder",
```
sosiouxme, line 274:
```suggestion
@click.option("--close-placeholder", "close_placeholder",
```

elliottlib/cli/__main__.py
```diff
@@ -96,8 +96,12 @@
 @click.option("--id", metavar='BUGID', default=[],
               multiple=True, required=True,
               help="Bugzilla IDs to add.")
+@click.option("--auto",
```
sosiouxme, line 99: i liked your `--all` idea better, but naming a var `all` could be bad... what about:
```suggestion
@click.option("--all", "remove_all",
```

sosiouxme (changes requested):
got a few problems but on the right path...

elliottlib/cli/__main__.py
```diff
@@ -362,6 +378,11 @@ def repair_bugs(runtime, advisory, auto, id, original_state, new_state, noop, de
         if bug.status in original_state:
             changed_bug_count += 1
             elliottlib.bzutil.set_state(bug, new_state, noop=noop)
+        if close_placeholder and "Placeholder" in bug.summary:
+            changed_bug_count += 1
+            elliottlib.bzutil.set_state(bug, "CLOSED", noop=noop)
```
vfreex, line 383: This would move placeholder bugs even if it is not in original_state.
Also for placeholder bugs, `changed_bug_count` may be added twice, and the state may be unnecessarily changed twice.
```suggestion
        if bug.status in original_state:
            if close_placeholder and "Placeholder" in bug.summary:
                elliottlib.bzutil.set_state(bug, "CLOSED", noop=noop)
            else:
                elliottlib.bzutil.set_state(bug, new_state, noop=noop)
            changed_bug_count += 1
```

elliottlib/cli/__main__.py
```diff
@@ -362,6 +378,11 @@ def repair_bugs(runtime, advisory, auto, id, original_state, new_state, noop, de
...
+        if comment and not noop:
+            bug.setstatus(comment=comment, private=False)
```
vfreex, line 385: This line can be moved to the scope where set_state is called, because we don't need to add a comment to the placeholder bug.

tests/test_errata.py
```diff
@@ -249,6 +249,9 @@ def commit(self):
     def addBugs(self, buglist):
         pass
 
+    def removebugs(self, buglist):
+        pass
```
vfreex, line 253: ?

elliottlib/cli/__main__.py
```diff
@@ -94,10 +94,14 @@
               help='Remove found bugs from ADVISORY')
 @use_default_advisory_option
 @click.option("--id", metavar='BUGID', default=[],
-              multiple=True, required=True,
+              multiple=True, required=False,
               help="Bugzilla IDs to add.")
+@click.option("--all", "remove_all",
+              required=False,
+              default=False, is_flag=True,
+              help="AUTO mode, remove all bugs attached to ADVISORY")
```
vfreex, line 102: `AUTO mode` sound vague to me.
would like to rephrase to something like `Remove all attached bugs from ADVISORY`

vfreex (changes requested):

elliottlib/cli/__main__.py
```diff
@@ -138,9 +141,15 @@ def remove_bugs(runtime, advisory, default_advisory_type, id):
...
+            if remove_all:
+                bug_ids = advs.errata_bugs
+            else:
+                bug_ids = [bzapi.getbug(i) for i in cli_opts.id_convert(id)]
+            green_prefix("Found {} bugs:".format(len(bug_ids)))
+            click.echo(" {}".format(", ".join([str(b) for b in bug_ids])))
             green_prefix("Removing {count} bugs from advisory:".format(count=len(bug_ids)))
             click.echo(" {advs}".format(advs=advisory))
-            advs.removeBugs([bug.id for bug in bug_ids])
+            advs.removeBugs([bug for bug in bug_ids])
```
vfreex, line 152: I think `removeBugs` only accept a list of bug IDs instead of the bug objects.

elliottlib/cli/__main__.py
```diff
@@ -138,9 +141,15 @@ def remove_bugs(runtime, advisory, default_advisory_type, id):
             raise ElliottFatalError("Error: Could not locate advisory {advs}".format(advs=advisory))
 
         try:
+            if remove_all:
+                bug_ids = advs.errata_bugs
+            else:
+                bug_ids = [bzapi.getbug(i) for i in cli_opts.id_convert(id)]
```
vfreex, line 147: The variable name is `bug_ids`, but they are actually bug objects.

elliottlib/cli/__main__.py
```diff
@@ -360,8 +376,14 @@ def repair_bugs(runtime, advisory, auto, id, original_state, new_state, noop, de
     green_print("Got bugs for advisory")
     for bug in attached_bugs:
         if bug.status in original_state:
+            if close_placeholder and "Placeholder" in bug.summary:
+                elliottlib.bzutil.set_state(bug, "CLOSED", noop=noop)
+            else:
+                elliottlib.bzutil.set_state(bug, new_state, noop=noop)
+                # only add comments for non-placeholder bug
+                if comment and not noop:
+                    bug.setstatus(comment=comment, private=False)
```
vfreex, line 385: The `else` branch will be reached if  `close_placeholder == False` or `"Placeholder" not in bug.summary`.

vfreex (changes requested):

PR #250, opened 2022-01-13
[ART-3665] flatten out settings file extensions iteration to avoid redundant cycles

locriandev:
I've used a list comprehension to derive this list: ['yaml', 'YAML', 'yml', 'YML', 'json', 'JSON'] from the original VALID_EXT list. The nested loop can be then flattened out.


elliottlib/dotconfig.py
```diff
@@ -57,17 +57,16 @@ def __init__(self, app=None, name=None, base_dir='~/.config/',
...
 
-            for ext in VALID_EXT:
-                for _ext in [ext, ext.upper()]:
-                    filename = '{}.{}'.format(name, _ext)
-                    cfg = os.path.join(self.base_dir, filename)
-                    if os.path.isfile(cfg):
-                        self.full_path = cfg
-                        self.filename = filename
-                        break
+            for ext in [ext for t in zip(VALID_EXT, [item.upper() for item in VALID_EXT]) for ext in t]:
```
vfreex, line 60: This doesn't seem to avoid redundant cycles :)
From the perspective of time complexity, both of the implementations are the same `O(n)`, where `n` is the size of `VALID_EXT`.
The original loop breaks earlier when it finds a match. However, this new implementation always loops `2 * n` times to compute the full list `[ext for t in zip(VALID_EXT, [item.upper() for item in VALID_EXT]) for ext in t]` regardless of the data.
If you really want to reduce the cycles of loop, I think we can use iterators. Something like this:
```suggestion
            for ext in itertools.chain(VALID_EXT, map(str.upper, VALID_EXT)):
```

vfreex (changes requested):

vfreex (approved):
Thanks.
/lgtm

PR #254, opened 2022-02-02
Batch bug lookup

thegreyd:


elliottlib/cli/find_bugs_cli.py
```diff
@@ -226,8 +225,13 @@ def find_bugs_cli(runtime: Runtime, advisory, default_advisory_type, mode, check
             sweep_cutoff_timestamp = bzutil.approximate_cutoff_timestamp(runtime.assembly_basis_event, brew_api, runtime.rpm_metas() + runtime.image_metas())
 
         if sweep_cutoff_timestamp:
-            green_print(f"Filtering bugs that have changed to one of the desired statuses before the cutoff time {datetime.utcfromtimestamp(sweep_cutoff_timestamp)}...")
-            qualified_bugs = bzutil.filter_bugs_by_cutoff_event(bzapi, bugs, status, sweep_cutoff_timestamp)
+            green_print(f"Filtering bugs that have changed ({len(bugs)}) to one of the desired statuses before the "
+                        f"cutoff time"
+                        f" {datetime.utcfromtimestamp(sweep_cutoff_timestamp)}...")
+            qualified_bugs = []
+            for chunk_of_bugs in chunk(bugs, 500):
```
locriandev, line 232: I always prefer explicit module level "constants" insted of hardcoded numeric values that will be hard to maintain in the future

elliottlib/util.py
```diff
@@ -687,3 +687,8 @@ def go_suffix_for_arch(arch: str) -> str:
 def brew_suffix_for_arch(arch: str) -> str:
     arch = brew_arch_for_go_arch(arch)  # translate either incoming arch style
     return brew_arch_suffixes[brew_arches.index(arch)]
+
+
+def chunk(l, n):
```
locriandev: How about adding type hints to this function?

locriandev: I would also pick more meaningful names for parameters: as I understand, "l" will always be a bug list while "n" is the batch size

thegreyd: Good point, added

PR #257, opened 2022-02-08
Exit code 1 prevents capturing stdout in Jenkins

locriandev:
Currently, Jenkins does not allow capturing the `stdout` of a failing command, even when the exception is caught. This prevents our automation from notifying the details of the issue encountered. Specifically, the `elliott verify-bugs` command output will not be available when regressions are found. I suggest we change this behavior until this bug is fixed in Jenkins:

https://issues.jenkins.io/browse/JENKINS-44930


locriandev:
Closing as this would break the logic for which `promote-assembly` stops when `verify-attached-bugs` fail

PR #259, opened 2022-02-14
[ART-3784] Add json format to --report option

locriandev:
Adding a --output option to find-bugs command that spits out JSON formatted output to stdout. This output will be then easily parsed by any job using this command, for example for customizable Slack messages.

### DIFF
--- a/elliottlib/cli/find_bugs_cli.py
+++ b/elliottlib/cli/find_bugs_cli.py
@@ -1,4 +1,6 @@
 import datetime
+import json
+
 from elliottlib.assembly import assembly_issues_config
 import re
 from datetime import datetime
@@ -65,6 +67,11 @@ LOGGER = logutil.getLogger(__name__)
               required=False,
               is_flag=True,
               help="Output a detailed report of the found bugs")
+@click.option('--output',
+              required=False,
+              type=click.Choice(['text', 'json']),
+              default='text',
+              help='Applies chosen format to --report output')
 @click.option("--into-default-advisories",
               is_flag=True,
               help='attaches bugs found to their correct default advisories, e.g. operator-related bugs go to "extras" instead of the default "image", bugs filtered into "none" are not attached at all.')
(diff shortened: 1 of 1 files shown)


elliottlib/cli/find_bugs_cli.py
```diff
@@ -65,6 +67,11 @@
               required=False,
               is_flag=True,
               help="Output a detailed report of the found bugs")
+@click.option('--output',
```
thegreyd, line 70: -o as a short flag would be nice

elliottlib/cli/find_bugs_cli.py
```diff
@@ -208,8 +215,9 @@ def find_bugs_cli(runtime: Runtime, advisory, default_advisory_type, mode, check
         if mode != 'qe' and exclude_status:
             status = set(status) - set(exclude_status)
 
-        green_prefix(f"Searching for bugs with status {' '.join(status)} and target release(s):")
-        click.echo(" {tr}".format(tr=", ".join(bz_data['target_release'])))
+        if output == 'text':
```
thegreyd, line 218: These ideally should be logger.info, suppressed by --quiet. Would leave it to you, if you want to do that here or not

locriandev: I'd rather not, this is a issue we might want to tackle but I would do it in a separate task. Would leave as-is for now

elliottlib/cli/find_bugs_cli.py
locriandev: Agree, added

elliottlib/cli/find_bugs_cli.py
```diff
@@ -397,19 +406,40 @@ def add_flags(bugs: type_bug_list, flags: List[str], noop: bool) -> None:
...
-                                                                                bug.component,
-                                                                                bug.status,
-                                                                                bug.cf_pm_score if hasattr(bug,
-                                                                                                           "cf_pm_score") else '?',
-                                                                                days_ago,
-                                                                                bug.summary[:60]))
+def print_report(bugs: type_bug_list, output: str = 'text') -> None:
+    if output == 'slack':
+        for bug in bugs:
+            click.echo("[{}]({:<12s}) {:<25s} ".format(bug.id, bug.weburl, bug.component))
```
thegreyd, line 412: Would the weburl be hyperlinked ? Would prefer that personally .. thoughts?

thegreyd: Oh it would be, I see it now

PR #260, opened 2022-02-16
verify-attached-bugs: Fix some --verify-flaws issues

vfreex:
- The Errata server doesn't seem to support keep-alive. Using `force_close=True` option to solve the `aiohttp.client_exceptions.ServerDisconnectedError`.
- Exit with non-zero if `--verify-flaws` finds any issues.


joepvd (approved):
/lgtm

PR #261, opened 2022-02-16
verify-attached-bugs: fix flaw bugs check issue when no trackers are attached

vfreex:
Fix error message like

```
Couldn't determine target release for attached trackers: Found different target_release values for tracker bugs: set(). There should be only 1 target release for all bugs. Fix the offending bug(s) and try again.
Some bug problems were listed above. Please investigate.
```

That error comes from current_target_release, err = util.get_target_release(attached_trackers) . The function raises an error if attached_trackers is empty array/set.


joepvd:
Looks even better

/lgtm

PR #267, opened 2022-03-07
Fixes for y-stream cve first fix script

sfowl:
Fixes two issues:
 - Lack of handling for BZ pagination
 - Incorrect sorting/comparsion of semver target releases

This standalone script is only really useful for human review of Y-stream errata, I'm not sure if the core elliotlib code may also have a similar issue with comparison of target releases.


vfreex (approved):
/lgtm

vfreex:
Though I didn't test this. Merging.

PR #279, opened 2022-03-29
Remove rpm-py-installer

vfreex:
This package is RHEL-only, which also caused a lot of build failures in
the past.

It doesn't seem to be actually used in Elliott.


elliottlib/rpm_utils.py
```diff
@@ -0,0 +1,68 @@
+from typing import Dict, Optional
```
joepvd, line 1: Might be nice with comment with a link to the doozer implementation would be helpful

joepvd (approved):
/lgtm

PR #285, opened 2022-04-06
[ART-3826] refactor create bug function

Ximinhan:
refactor create bug function used by create placeholder bug and create text_only bug

### DIFF
--- a/elliottlib/bzutil.py
+++ b/elliottlib/bzutil.py
@@ -82,7 +82,7 @@ class BugzillaBugTracker:
             id_bug_map[ids[i]] = bug
         return id_bug_map
 
-    def create_bug(self, bugtitle, bugdescription, keywords: List):
+    def create_bug(self, bugtitle, bugdescription, target_status, keywords: List):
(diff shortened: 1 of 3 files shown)


elliottlib/bzutil.py
```diff
@@ -82,6 +82,33 @@ def get_bugs_map(self, ids: List[int], raise_on_error: bool = True, **kwargs) ->
...
+            product=self.config.get('product'),
+            version=self.config.get('version')[0],
+            component="Release",
+            summary=bugtitle,
+            keywords=keywords,
+            description=bugdescription)
+        newbug = self._client.createbug(createinfo)
+        # change state to VERIFIED, set target release
+        try:
+            update = self._client.build_update(status="VERIFIED", target_release=self.config.get('target_release'))
```
thegreyd, line 96: Can you accept a state, that way this method would be more general. And then pass VERIFIED from placeholder/textonly methods

Ximinhan: updated

PR #307, opened 2022-04-19
[ART-3825] refactor set_state function in bzutil

Ximinhan:
add update_bug_status and add_comment_to_bug function for bugtracker to replace set_state function


elliottlib/bzutil.py
```diff
@@ -177,8 +177,22 @@ def get_remotelinks(self, bugid):
                 linkdict[link.relationship] = link.object.url
         return linkdict
 
-    def update_bug(self, bugid, status):
-        self._client.transition_issue(self._client.issue(bugid), status)
+    def update_bug_status(self, bugid, target_status, comment_for_release=None, noop=False):
+        current_state = self._client.issue(bugid).fields.status.name
+        if noop:
+            logger.info(f"Would have changed BZ#{bugid} from {current_state} to {target_status}")
```
thegreyd, line 183: Let's not mention "BZ" here

elliottlib/bzutil.py
```diff
@@ -177,8 +177,22 @@ def get_remotelinks(self, bugid):
...
+            return
+        self._client.transition_issue(self._client.issue(bugid), target_status)
+        comment = f'Elliott changed bug status from {current_state} to {target_status}.'
+        self.add_comment_to_bug(self, bugid, comment, True, comment_for_release)
```
thegreyd, line 187: I would like this to have a single optional param "comment" for custom comments, and then we can have another param called "status_update_comment=True" for the standard comment message. 

And then let's make add_comment_to_bug a generic/dumb method which will just add a comment string passed to it. Then we have 2 calls to comment_to_bug from this method

elliottlib/bzutil.py
```diff
@@ -177,8 +177,22 @@ def get_remotelinks(self, bugid):
                 linkdict[link.relationship] = link.object.url
         return linkdict
 
-    def update_bug(self, bugid, status):
-        self._client.transition_issue(self._client.issue(bugid), status)
+    def update_bug_status(self, bugid, target_status, comment_for_release=None, noop=False):
+        current_state = self._client.issue(bugid).fields.status.name
```
thegreyd, line 181: Let's make this an attribute (self.status) so we can easily fetch it

elliottlib/bzutil.py
```diff
@@ -177,8 +177,22 @@ def get_remotelinks(self, bugid):
                 linkdict[link.relationship] = link.object.url
         return linkdict
 
-    def update_bug(self, bugid, status):
-        self._client.transition_issue(self._client.issue(bugid), status)
+    def update_bug_status(self, bugid, target_status, comment_for_release=None, noop=False):
```
thegreyd, line 180: I would actually like these methods to be in JIRABug and BugzillaBug classes
so you can just call `bug.update_status()` and `bug.add_comment()`

elliottlib/bzutil.py
```diff
@@ -301,9 +315,22 @@ def create_bug(self, bugtitle, bugdescription, target_status, keywords: List):
             sleep(5)
             self._client.update_bugs([newbug.id], update)
             print(ex)
-
         return newbug
 
+    def update_bug_status(self, bugid, target_status, comment_for_release=None, noop=False):
```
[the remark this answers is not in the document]
Ximinhan: it's for different bug tracker

elliottlib/bzutil.py
Ximinhan: for bugzilla bug it can work as it used to be, but for jirabug, update status need init jira client, it's not a self function for jira issue

PR #310, opened 2022-04-20
set target_release when create bug

Ximinhan:
we can also set status to VERIFIED when create bug

### DIFF
--- a/elliottlib/bzutil.py
+++ b/elliottlib/bzutil.py
@@ -295,6 +295,7 @@ class BugzillaBugTracker(BugTracker):
         create_info = self._client.build_createbug(
             product=self.config.get('product'),
             version=self.config.get('version')[0],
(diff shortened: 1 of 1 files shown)


elliottlib/bzutil.py
```diff
@@ -295,14 +295,15 @@ def create_bug(self, title, description, target_status, keywords: List) -> Bugzi
...
+            target_release=self.config.get('target_release')[0],
             component="Release",
             summary=title,
             keywords=keywords,
             description=description)
         new_bug = self._client.createbug(create_info)
         # change state to VERIFIED, set target release
         try:
-            update = self._client.build_update(status=target_status, target_release=self.config.get('target_release')[0])
+            update = self._client.build_update(status=target_status)
```
vfreex, line 306: The comment at line 304 also needs an update. otherwise LGTM.

PR #311, opened 2022-04-20
find-builds: Exclude builds that have been attached

vfreex:
Current behavior:
Only builds that are attached to other advisories that belong to the same
major.minor version will be excluded.

This PR:
Builds that are attached to any advisories will be excluded.

### DIFF
--- a/elliottlib/cli/find_builds_cli.py
+++ b/elliottlib/cli/find_builds_cli.py
@@ -181,7 +181,7 @@ PRESENT advisory. Here are some examples:
             lambda nvrp: elliottlib.errata.get_brew_build('{}-{}-{}'.format(nvrp[0], nvrp[1], nvrp[2]), nvrp[3], session=requests.Session())
         )
         if not allow_attached:
-            unshipped_builds = _filter_out_inviable_builds(kind, unshipped_builds, elliottlib.errata)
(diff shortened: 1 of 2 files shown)


elliottlib/cli/find_builds_cli.py
```diff
@@ -414,27 +414,6 @@ def _fetch_builds_by_kind_rpm(runtime: Runtime, tag_pv_map: Dict[str, str], brew
...
-                # though not very useful (there's a command for adding them,
-                # but not much point in doing it). just looking at the first one is fine.
-                errata_version_cache[eid] = metadata_comments_json[0]['release']
-            if errata_version_cache[eid] == get_release_version(b.product_version):
-                in_same_version = True
-                break
-        if not in_same_version:
-            unshipped_builds.append(b)
-    return unshipped_builds
+def _filter_out_attached_builds(kind, results, errata):
```
thegreyd, line 417: let's remove the kind param?

sosiouxme (changes requested):
i think there's still a need to attach the same RPM builds to multiple releases occasionally for our BYO RHEL users. this returns to our unresolved "what about inheritance" thread but even with inheritance, we could easily have a situation where the same RPM needed a security update in multiple releases and had to actually ship in all. so i think we still need a way to distinguish which advisories belong to which releases and which builds have shipped in the same release.

PR #314, opened 2022-04-20
[ART-3981] Improve version handling for Elliott

locriandev:
Tested a local build:
```
python3 setup.py bdist_wheel --universal
python3 -m twine check dist/*.whl
pip install dist/*.whl
```
`elliott --version` prints:
```
Elliott v2.0.12-8-g7e0b9b8
```


setup.py
```diff
@@ -1,32 +1,44 @@
...
-    version=_get_version(),
+    version=__version__,
     description="CLI tool for managing and automating Red Hat software releases",
     long_description=open('README.md').read(),
     long_description_content_type="text/markdown",
     url="https://github.com/openshift/elliott",
     license="Apache License, Version 2.0",
     packages=find_packages(exclude=["tests", "tests.*", "functional_tests", "functional_tests.*"]),
     include_package_data=True,
+    package_data={'': ['elliottlib/VERSION']},
```
vfreex, line 41: We have `MANIFEST.in`. `package_data` shouldn't be used together.

vfreex (changes requested):

setup.py
locriandev: fixed

elliottlib/__init__.py
```diff
@@ -1,10 +1,11 @@
...
 if sys.version_info < (3, 6):
     sys.exit('Sorry, Python < 3.6 is not supported.')
 from .runtime import Runtime
 
 
 def version():
-    from os.path import abspath, dirname, join
-    filename = join(dirname(abspath(__file__)), 'VERSION')
-    return open(filename).read().strip()
+    stream = pkg_resources.resource_stream('elliottlib', 'VERSION')
```
vfreex, line 10: I think we also need the ability to run elliott locally. So if that file doesn't exist, we need to fallback to the `git describe` approach.

setup.py
```diff
@@ -1,25 +1,36 @@
...
 with open('./requirements.txt') as f:
     INSTALL_REQUIRES = f.read().splitlines()
 
 
-def _get_version():
-    from os.path import abspath, dirname, join
-    filename = join(dirname(abspath(__file__)), 'elliottlib', 'VERSION')
-    return open(filename).read().strip()
+def get_version():
+    proc = subprocess.run(['git', 'describe', '--tags'], stdout=subprocess.PIPE)
```
vfreex, line 14: Add `check=True`.

elliottlib/__init__.py
locriandev: fixed, good catch

setup.py
vfreex: I didn't see it was added.

elliottlib/__init__.py
```diff
@@ -1,10 +1,16 @@
...
 
 def version():
-    from os.path import abspath, dirname, join
-    filename = join(dirname(abspath(__file__)), 'VERSION')
-    return open(filename).read().strip()
+    try:
+        stream = pkg_resources.resource_stream('elliottlib', 'VERSION')
+        return stream.read().decode()
+    except FileNotFoundError:
+        proc = subprocess.run(['git', 'describe', '--tags'], stdout=subprocess.PIPE)
```
vfreex, line 15: same here.

locriandev:
closed in favor of https://github.com/openshift/elliott/pull/319

PR #315, opened 2022-04-20
[ART-3823] add blocks and version for jirabug

Ximinhan:


elliottlib/bzutil.py
```diff
@@ -100,11 +100,21 @@ def __init__(self, bug_obj: Issue):
...
         self.summary = self.bug.fields.summary
+        self.blocks = self.get_blocks()
+        self.keywords = self.bug.fields.labels
+        self.version = [x.name for x in self.bug.fields.versions]
         self.target_release = [x.name for x in self.bug.fields.fixVersions]
 
     def creation_time_parsed(self):
         return datetime.strptime(str(self.bug.fields.created), '%Y-%m-%dT%H:%M:%S.%f%z')
 
+    def get_blocks(self):
```
thegreyd, line 111: let's rename this to _get_blocks() since it's an internal func

elliottlib/bzutil.py
```diff
@@ -100,11 +100,21 @@ def __init__(self, bug_obj: Issue):
         self.component = self.bug.fields.components[0].name
         self.status = self.bug.fields.status.name
         self.summary = self.bug.fields.summary
+        self.blocks = self.get_blocks()
+        self.keywords = self.bug.fields.labels
```
thegreyd, line 104: I'm not sure about this change.. keywords are going to go away when bugzilla goes away.
But for now I think it's okay

PR #319, opened 2022-04-21
[ART-3981] Improve version handling for Elliott

locriandev:
The proposed approach is to use [setuptools-scm](https://github.com/pypa/setuptools_scm) package to have a unique version id for each git commit. This will avoid confusion by making it clear which tool state we are actually running.

If the current commit has a tag, let's say `v2.0.13`, running `elliott --version` will print `Elliott v2.0.13`. If last tag happened 8 commits ago and was named `v2.0.12`, `elliott --version` will print `Elliott v2.0.13.dev8+gfd18ab5` where `fd18ab5` is the short  version hash of the current (untagged) commit.

This will be reflected in the `rh-elliott` package being installed by running `setup.py`, so Jenkins console output will make it clear the version that's being used within a given pipeline.


Jenkinsfile
```diff
@@ -5,7 +5,7 @@ pipeline {
         docker {
             image "openshift-art/art-ci-toolkit:latest"
             alwaysPull true
-            args "-e http_proxy -e https_proxy -e no_proxy -e HTTP_PROXY -e HTTPS_PROXY -e NO_PROXY --entrypoint=''"
+            args "-v /home/jenkins/.ssh:/home/art/.ssh -e http_proxy -e https_proxy -e no_proxy -e HTTP_PROXY -e HTTPS_PROXY -e NO_PROXY --entrypoint=''"
```
vfreex, line 8: Could you explain why `.ssh` directory is required?

vfreex (commented):
Could you explain why .ssh directory is required?
Otherwise I'll LGTM.

Jenkinsfile
locriandev: I initially didn't manage to have Jenkins fetch tags from remote (tags are required by the new versioning scheme). Setting "Advanced clone behaviours - Fetch tags" flag in [job configuration](https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/art-tools/configure) did the job. Manual fetching can be removed, ssh keys bind mount is no longer necessary

PR #322, opened 2022-04-22
[ART-3821] attach jira cve flaws

Ximinhan:


Ximinhan:
waiting on `resolution` and `alias` and `whiteboard` fields get confirmed in jira

PR #326, opened 2022-04-27
fix create_bug arg

Ximinhan:


elliottlib/bzutil.py
```diff
@@ -365,6 +365,9 @@ def create_bug(self, title, description, target_status, keywords: List) -> Bugzi
             keywords=keywords,
             description=description)
         new_bug = self._client.createbug(create_info)
+        if noop:
```
vfreex, line 368: Seems to be too late to check noop. Bug is already created when this line is reached.

vfreex (changes requested):

vfreex (approved):
I am not sure if returning None will cause other issues. I am going to merge this PR but l will leave the question to @Ximinhan to investigate.

PR #331, opened 2022-04-28
Fix bug categorization issue with sub_components

vfreex:
`sub_components` field is not in the bug query so `bug.sub_component` field is absent.


joepvd:
Nice find! Think it would not hurt to have that attribute also added to `def blocker_search`

sosiouxme:
/approve
fix the tests tho :-P

thiagoalessio (approved):
/lgtm

Ximinhan:
rebased in https://github.com/openshift/elliott/pull/337, close this for now

PR #349, opened 2022-06-13
[ART-4152] use errata api remove bug

Ximinhan:
use errata api to drop bug instead of using Erratum, Erratum drop bug will failed on last bug post
```
$ python3 ./elliott --group=openshift-4.10 remove-bugs --advisory 97030 --all 
2022-06-16 16:01:56,927 INFO Cloning config data from https://github.com/openshift/ocp-build-data.git 
2022-06-16 16:01:59,321 INFO Using branch from group.yml: rhaos-4.10-rhel-8 
[2097429] 
Found 1 bugs: 2097429 
Removing bugs from advisory 97030:⏎ 
```


elliottlib/errata.py
```diff
@@ -125,6 +125,22 @@ def remove_multi_jira_issues(advisory_id, jira_list: List):
     return [remove_jira_issue(advisory_id, jira_id) for jira_id in jira_list]
 
 
+def remove_bug(advisory_id, bug_id):
+    """
+    Remove a bug from advisory
+    Response code will return
+    """
+    return ErrataConnector()._post(f"/api/v1/erratum/{advisory_id}/remove_bug", data={"bug": f"{bug_id}"})
```
vfreex, line 133: Using internal methods (starting with `_`) is not recommended. It is more likely to be broken.

elliottlib/errata.py
```diff
@@ -125,6 +125,22 @@ def remove_multi_jira_issues(advisory_id, jira_list: List):
...
+
+
+def remove_multi_bugs(advisory_id, bug_list: List):
+    """
+    Remove multi bugs from advisory
+    Return a list of response code
+    """
+    return [remove_bug(advisory_id, bug_id) for bug_id in bug_list]
```
vfreex, line 141: This is not performant. For N bugs, it will create N `ErrataConnector` instances. Each will create a new TCP connection and kerberos authentication.

Ximinhan: yes, but errata don't have batch remove / add api, same for jira issues

thegreyd: I think what yuxiang is suggesting is to use the same ErrataConnector instance for each request

thegreyd:
@Ximinhan I see in the story that this is more of an edge case with side effect - i.e. happens only when there's a single bug and then it gets removed, so I'd suggest we ignore the error in this specific case

Ximinhan:
I think it's happened on processing the last bug, ref https://github.com/red-hat-storage/errata-tool/blob/master/errata_tool/erratum.py#L892-L895
considering when we drop a release, we need to drop extra and olm advisory, they usually only have one placeholder bug, so I guess we will meet it again

PR #353, opened 2022-06-15
remove-bugs fixes

thegreyd:
Note: errata api for removing jira bugs is broken. I'll follow up with the team

Summary
- Make sure bug id given is attached to the advisory
- Abstract out out bug remove logic into bug tracker methods
- Change the cli interface (no longer `--id`) to be consistent with attach-bugs

Test
- `make test`
- `elliott -g openshift-4.10 --assembly 4.10.19 remove-bugs --all --use-default-advisory image --noop`
- `elliott -g openshift-4.10 --assembly 4.10.19 remove-bugs 2095114 --use-default-advisory image --noop`
- `USEJIRA=True ./elliott -g openshift-4.10 remove-bugs --advisory 96435 OCPBUGS-1`

### DIFF
--- a/elliottlib/cli/remove_bugs_cli.py
+++ b/elliottlib/cli/remove_bugs_cli.py
@@ -34,13 +34,17 @@ def remove_bugs_cli(runtime, advisory_id, default_advisory_type, bug_ids, remove
     Remove bugs that have been attached an advisory:
 
 \b
-    $ elliott --group openshift-3.7 remove-bugs --id 123456 --advisory 1234123
+    $ elliott --group openshift-4.10 remove-bugs 123456 --advisory 1234123
 
-    Remove two bugs from default rpm advisory. Note that --group is required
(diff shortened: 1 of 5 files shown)


elliottlib/cli/remove_bugs_cli.py
```diff
@@ -53,55 +42,50 @@ def remove_bugs_cli(runtime, advisory, default_advisory_type, id, remove_all):
...
+                    bug_tracker, noop)
+
+
+def remove_bugs(advisory_id, bug_ids, remove_all, bug_tracker, noop):
+    try:
+        advisory = errata.Advisory(errata_id=advisory_id)
+    except GSSError:
+        exit_unauthenticated()
+
+    if advisory is False:
```
ashwindasr, line 77: `if not advisory` could be a alternative for `if advisory is False`

PR #357, opened 2022-06-17
Make config backward compatible + regex fix

thegreyd:


functional_tests/test_find_bugs_blocker.py
```diff
@@ -8,14 +8,11 @@ def test_find_bugs_blocker(self):
         out = subprocess.check_output(
             constants.ELLIOTT_CMD
             + [
-                "--assembly=stream", "--group=openshift-4.3", "find-bugs:blocker", '--exclude-status=ON_QA'
+                "--assembly=stream", "--group=openshift-4.6", "find-bugs:blocker", '--exclude-status=ON_QA'
```
joepvd, line 11: Could we take the most recent future EUS release (4.10 i think)? Then we can forget about this the longest time :)

joepvd:
/lgtm

Left a comment, but that is optional :)

functional_tests/test_find_bugs_blocker.py
sosiouxme: agreed but i do not want to hold this PR waiting on that...

sosiouxme (approved):
/lgtm

PR #364, opened 2022-07-01
Sweep TALM && BMER Bugs

thiagoalessio:
With the addition of new images on 4.10 (ART-4052, ART-4074, ART-4075, and ART-4076) we also need to sweep their corresponding bugs into advisories.


thiagoalessio:
/hold

I want this to be true for 4.10, but not for other versions ... how can we achieve that?

Ximinhan:
this is just distribute bug to extras, whether check it defined in ocp-build-data bug.yml

thiagoalessio:
/hold cancel

oh, makes sense! thanks @Ximinhan !

PR #366, opened 2022-07-14
[ART-4172] handle non-visible tracker bugs in first fix

Ximinhan:


elliottlib/bzutil.py
```diff
@@ -1063,8 +1063,17 @@ def is_first_fix_any(bugtracker, flaw_bug, current_target_release):
         return True
 
     # filter tracker bugs by OCP product
-    tracker_bugs = [b for b in bugtracker.get_bugs(tracker_ids)
-                    if b.product == constants.BUGZILLA_PRODUCT_OCP and b.is_tracker_bug()]
+    tracker_bugs = []
+    for tracker_id in tracker_ids:
+        try:
+            b = bugtracker.get_bug(tracker_id)
+        except Exception as e:
```
locriandev, line 1070: Is there a less generic Exception we can catch?

elliottlib/bzutil.py
```diff
@@ -1063,8 +1063,17 @@ def is_first_fix_any(bugtracker, flaw_bug, current_target_release):
...
+            # first-fix tracker bug might be not visible but not break here
+            logger.warn(e)
```
locriandev, line 1072: I think we could use a more explanatory warning message to grep the logs for when needed

locriandev (changes requested):

elliottlib/bzutil.py
Ximinhan: there is https://github.com/openshift/elliott/blob/master/elliottlib/exceptions.py#L31 but seems just a mask

elliottlib/bzutil.py
Ximinhan: as previous, I can't distinguish the unauthorized error, so print all message here

elliottlib/bzutil.py
```diff
@@ -1063,8 +1063,18 @@ def is_first_fix_any(bugtracker, flaw_bug, current_target_release):
...
     # filter tracker bugs by OCP product
-    tracker_bugs = [b for b in bugtracker.get_bugs(tracker_ids)
-                    if b.product == constants.BUGZILLA_PRODUCT_OCP and b.is_tracker_bug()]
+    tracker_bugs = []
+    for tracker_id in tracker_ids:
+        try:
+            b = bugtracker.get_bug(tracker_id)
+        except Exception as e:
+            # first-fix tracker bug might be not visible but not break here
+            logger.info("Failed to get tracker bug info from bugzilla API, need manually check if it's permission issue")
```
vfreex, line 1072: `info` is not obvious to release artists. It should be a warning.
Also please include the bug ID and the original exception message in the log message.

elliottlib/bzutil.py
```diff
@@ -1063,8 +1063,18 @@ def is_first_fix_any(bugtracker, flaw_bug, current_target_release):
...
+            logger.warning(e)
```
vfreex, line 1073: Please write a log message instead of a string of the exception object.

elliottlib/bzutil.py
```diff
@@ -1063,8 +1063,18 @@ def is_first_fix_any(bugtracker, flaw_bug, current_target_release):
         return True
 
     # filter tracker bugs by OCP product
-    tracker_bugs = [b for b in bugtracker.get_bugs(tracker_ids)
-                    if b.product == constants.BUGZILLA_PRODUCT_OCP and b.is_tracker_bug()]
+    tracker_bugs = []
+    for tracker_id in tracker_ids:
+        try:
+            b = bugtracker.get_bug(tracker_id)
```
vfreex, line 1069: Getting bugs one by one could be slow. Is `get_bugs` not possible to skip certain individual errors?

vfreex (changes requested):

elliottlib/bzutil.py
Ximinhan: the authorized failure happened in getbugs functions from bugzilla lib, I don't think it could be modified

elliottlib/bzutil.py
Ximinhan: updated

elliottlib/bzutil.py
locriandev: I think this can still be improved: https://github.com/openshift/elliott/pull/366#discussion_r921000858

PR #367, opened 2022-07-27
Halt if rpm CVEs don't have corresponding build

joepvd:
Throw an error when there are CVE tracker bugs that cannot get attached
because the component that they are about is not getting shipped.

Currently an event like this will go unnoticed, as the log is not
scrutinized.

The can be bypassed by either fixing the issue, or by exclkuding the
relevant bugs from the assembly:

`releases.<name>.issues.exclude: [1234]`


thiagoalessio (approved):
/lgtm

sosiouxme (approved):
it would be nice if these could be collected, in case there are multiple, before halting execution. however that situation is probably rare enough, and a minor enough annoyance (doing multiple iterations instead of just one) that it's probably not worth complicating the logic.

PR #374, opened 2022-08-09
Run attach-cve-flaws for all advisores for an assembly

thegreyd:
This is mostly for release artists running manually.
`elliott --assembly=4.11.0 --group=openshift-4.11 attach-cve-flaws --into-default-advisories --noop`


vfreex (approved):
/lgtm

PR #393, opened 2022-08-22
Make use_jira default true

thegreyd:
Make usejira=True by default, so when we run elliott with no arguments/env vars
we will be using jira and bugzilla both by default in find commands (find-bugs and attach-cve-flaws)

Deprecate `ONLYJIRA` since it's creating confusion due to some commands only being able to run
in both bugzilla and jira mode, so supporting this is not guaranteed.

For commands which accept bug ids (attach-bugs, remove-bugs, repair-bugs, verify-bugs)
- To use Bugzilla ids with them `USEJIRA=0` or `USEJIRA=False`

### DIFF
--- a/elliottlib/cli/remove_bugs_cli.py
+++ b/elliottlib/cli/remove_bugs_cli.py
@@ -59,7 +59,7 @@ def remove_bugs_cli(runtime, advisory_id, default_advisory_type, bug_ids, remove
         for b in [runtime.bug_trackers('jira'), runtime.bug_trackers('bugzilla')]:
             remove_bugs(advisory_id, bug_ids, remove_all, b, noop)
     else:
-        if runtime.use_jira or runtime.only_jira:
+        if runtime.use_jira:
             bug_tracker = runtime.bug_trackers('jira')
         else:
             bug_tracker = runtime.bug_trackers('bugzilla')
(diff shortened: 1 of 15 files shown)


elliottlib/cli/remove_bugs_cli.py
```diff
@@ -55,12 +55,13 @@ def remove_bugs_cli(runtime, advisory_id, default_advisory_type, bug_ids, remove
...
 
-    if runtime.use_jira or runtime.only_jira:
-        bug_tracker = runtime.bug_trackers['jira']
-    else:
-        bug_tracker = runtime.bug_trackers['bugzilla']
-    bug_ids = bug_tracker.id_convert(bug_ids)
-    remove_bugs(advisory_id, bug_ids, remove_all, bug_tracker, noop)
+    if not remove_all and runtime.use_jira:
+        runtime.use_jira = False
+        runtime.only_jira = True
```
vfreex, line 60: This looks confusing to dynamically change those flags in runtime. Could you add a comment to explain why it is needed?

thegreyd: Good point :+1:

elliottlib/runtime.py
```diff
@@ -276,15 +277,15 @@ def image_metas(self):
...
-    @property
-    def bug_trackers(self):
-        if self._bug_trackers:
-            return self._bug_trackers
-        if not self.only_jira:
-            self._bug_trackers['bugzilla'] = BugzillaBugTracker(BugzillaBugTracker.get_config(self))
-        if self.use_jira or self.only_jira:
-            self._bug_trackers['jira'] = JIRABugTracker(JIRABugTracker.get_config(self))
-        return self._bug_trackers
+    def bug_trackers(self, bug_tracker_type):
```
thegreyd, line 280: This was a problematic piece of logic which was too rigid
for supporting all of different command configurations. so refactor
it to give it what the command wants

PR #399, opened 2022-08-24
Account for jira statuses + run verify-bugs only once

thegreyd:


elliottlib/cli/verify_attached_bugs_cli.py
```diff
@@ -69,9 +69,10 @@ async def verify_attached_bugs(runtime: Runtime, verify_bug_status: bool, adviso
 @click_coroutine
 async def verify_bugs_cli(runtime, verify_bug_status, output, bug_ids):
     runtime.initialize()
-    if runtime.use_jira:
+    if runtime.use_jira or runtime.only_jira:
```
vfreex, line 72: i think we should ensure `use_jira` is `True` if `only_jira` is `True`, otherwise we need to check both variables everywhere we need to check `use_jira`, which is easy to forget.
Maybe we can do that in `Runtime`?

vfreex (commented):
one minor issue, otherwise LGTM

locriandev:
Merging this as we need a quick fix to our alert messages in #forum-release

PR #402, opened 2022-08-25
Commands can recognize different bugids

thegreyd:
This change is for elliott commands that accept bug_ids as argument. 
Right now we need usejira=True to tell elliott that we are passing in 
jira ids and usejira=False to tell those are bugzilla ids. With this change
we try to be smart and automatically recognize bug_ids into jira and bz
and run accordingly. If pattern matching on bugids fails we crash.

Note: verify-bugs needs a refactor and will need follow up work before
having this change in.

## Testing
- `make test`
- `elliott -g openshift-4.9 attach-bugs [OCPBUGS-1](https://issues.redhat.com//browse/OCPBUGS-1) 2053222 -a 100250 --noop`
- `elliott -g openshift-4.9 remove-bugs [OCPBUGS-1](https://issues.redhat.com//browse/OCPBUGS-1) 2053222 -a 100250 --noop`

### DIFF
--- a/elliottlib/bzutil.py
+++ b/elliottlib/bzutil.py
@@ -31,6 +31,13 @@ def datetime_now():
     return datetime.now(timezone.utc)
 
 
+def get_jira_bz_bug_ids(bug_ids):
+    ids = cli_opts.id_convert_str(bug_ids)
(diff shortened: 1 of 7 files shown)


thegreyd:
Merging as this is blocking followup work

PR #403, opened 2022-08-26
Refactor and fix verify-attached-bugs 

thegreyd:
Status of refactor
- [X] get_attached_bugs
- [X] get_blocking_bugs_for
- [x] verify_attached_flaws_for

### DIFF
--- a/elliottlib/cli/verify_attached_bugs_cli.py
+++ b/elliottlib/cli/verify_attached_bugs_cli.py
@@ -3,6 +3,7 @@ import re
 from typing import Any, Dict, Iterable, List, Set, Tuple
 import click
 from spnego.exceptions import GSSError
+from errata_tool import Erratum
 
 
 from elliottlib import bzutil, constants, util, errata
@@ -45,7 +46,7 @@ async def verify_attached_bugs(runtime: Runtime, verify_bug_status: bool, adviso
     validator = BugValidator(runtime, use_jira, output="text")
     try:
         await validator.errata_api.login()
-        advisory_bugs = await validator.get_attached_bugs(advisories)
+        advisory_bugs = validator.get_attached_bugs(advisories)
         non_flaw_bugs = validator.filter_bugs_by_product({b for bugs in advisory_bugs.values() for b in bugs})
         validator.validate(non_flaw_bugs, verify_bug_status)
         if verify_flaws:
@@ -197,21 +198,19 @@ class BugValidator:
         if missing_cves:
             self._complain(f"On advisory {advisory_id}, bugs for the following CVEs are not attached but listed in advisory's `CVE Names` field: {', '.join(sorted(missing_cves))}")
 
-    async def get_attached_bugs(self, advisory_ids: Iterable[str]) -> Dict[int, Set[Bug]]:
+    def get_attached_bugs(self, advisory_ids: Iterable[str]) -> Dict[int, Set[Bug]]:
         """ Get bugs attached to specified advisories
         :return: a dict with advisory id as key and set of bug objects as value
         """
-        green_print(f"Retrieving bugs for advisory {advisory_ids}")
-        if self.use_jira:
-            issue_keys = {advisory_id: [issue["key"] for issue in errata.get_jira_issue_from_advisory(advisory_id)] for advisory_id in advisory_ids}
-            bug_map = self.bug_tracker.get_bugs_map([key for keys in issue_keys.values() for key in keys])
-            result = {advisory_id: {bug_map[key] for key in issue_keys[advisory_id]} for advisory_id in advisory_ids}
-        else:
-            advisories = await asyncio.gather(*[self.errata_api.get_advisory(advisory_id) for advisory_id in advisory_ids])
-            bug_map = self.bug_tracker.get_bugs_map(list({b["bug"]["id"] for ad in advisories for b in ad["bugs"]["bugs"]}))
-            result = {ad["content"]["content"]["errata_id"]: {bug_map[b["bug"]["id"]] for b in ad["bugs"]["bugs"]} for ad
-                      in advisories}
-        return result
+        green_print(f"Retrieving bugs for advisories: {advisory_ids}")
+        advisories = [Erratum(errata_id=advisory_id) for advisory_id in advisory_ids]
+        advisory_bug_id_map = {advisory.errata_id: self.bug_tracker.advisory_bug_ids(advisory)
+                               for advisory in advisories}
+        bug_map = self.bug_tracker.get_bugs_map([bug_id for bug_list in advisory_bug_id_map.values()
+                                                 for bug_id in bug_list])
+        attached_bug_map = {advisory_id: {bug_map[bid] for bid in advisory_bug_id_map[advisory_id]}
+                            for advisory_id in advisory_ids}
+        return attached_bug_map
 
     def filter_bugs_by_product(self, bugs):
         # filter out bugs for different product (presumably security flaw bugs)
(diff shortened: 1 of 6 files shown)


elliottlib/cli/verify_attached_bugs_cli.py
```diff
@@ -37,18 +38,23 @@ async def verify_attached_bugs_cli(runtime: Runtime, verify_bug_status: bool, ad
         exit(1)
     if runtime.use_jira:
         await verify_attached_bugs(runtime, verify_bug_status, advisories, verify_flaws, True)
-    await verify_attached_bugs(runtime, verify_bug_status, advisories, verify_flaws, False)
+    else:
```
joepvd, line 41: i think it should query bugzilla always.

elliottlib/cli/verify_attached_bugs_cli.py
```diff
@@ -37,18 +38,23 @@ async def verify_attached_bugs_cli(runtime: Runtime, verify_bug_status: bool, ad
...
+        await verify_attached_bugs(runtime, verify_bug_status, advisories, verify_flaws, False)
 
 
 async def verify_attached_bugs(runtime: Runtime, verify_bug_status: bool, advisories: Tuple[int, ...], verify_flaws: bool, use_jira: bool):
     validator = BugValidator(runtime, use_jira, output="text")
     try:
         await validator.errata_api.login()
-        advisory_bugs = await validator.get_attached_bugs(advisories)
-        non_flaw_bugs = validator.filter_bugs_by_product({b for bugs in advisory_bugs.values() for b in bugs})
+        advisory_bug_map = validator.get_attached_bugs(advisories)
```
joepvd, line 49: any reason await disappeared here?

joepvd (commented):
direction looks promising! :)

elliottlib/cli/verify_attached_bugs_cli.py
thegreyd: Yes this part isn't complete yet - use_jira will not be an option anywhere. We will by default be searching everywhere

elliottlib/cli/verify_attached_bugs_cli.py
thegreyd: Yes earlier we were doing async calls in get_attached_bugs but now we are not. I'll explain my reasoning as to why in the method

elliottlib/cli/verify_attached_bugs_cli.py
```diff
@@ -138,29 +151,35 @@ async def _verify_attached_flaws_for(self, advisory_id: int, attached_trackers:
...
                 first_fix_flaw_ids = {
                     flaw_bug.id for flaw_bug in flaw_id_bugs.values()
-                    if bzutil.is_first_fix_any(self.bug_tracker.client(), flaw_bug, current_target_release)
+                    if bzutil.is_first_fix_any(self.runtime.bug_trackers('bugzilla'), flaw_bug, current_target_release)
                 }
 
         # Check if attached flaws match attached trackers
         attached_flaw_ids = {b.id for b in attached_flaws}
-        missing_flaw_ids = attached_flaw_ids - first_fix_flaw_ids
+        missing_flaw_ids = first_fix_flaw_ids - attached_flaw_ids
```
thegreyd, line 159: I found this reversed for some reason @vfreex can you take a look I have it right?

elliottlib/cli/verify_attached_bugs_cli.py
```diff
@@ -180,41 +199,46 @@ async def _verify_attached_flaws_for(self, advisory_id: int, attached_trackers:
...
-        if self.use_jira:
-            issue_keys = {advisory_id: [issue["key"] for issue in errata.get_jira_issue_from_advisory(advisory_id)] for advisory_id in advisory_ids}
-            bug_map = self.bug_tracker.get_bugs_map([key for keys in issue_keys.values() for key in keys])
-            result = {advisory_id: {bug_map[key] for key in issue_keys[advisory_id]} for advisory_id in advisory_ids}
-        else:
-            advisories = await asyncio.gather(*[self.errata_api.get_advisory(advisory_id) for advisory_id in advisory_ids])
-            bug_map = self.bug_tracker.get_bugs_map(list({b["bug"]["id"] for ad in advisories for b in ad["bugs"]["bugs"]}))
-            result = {ad["content"]["content"]["errata_id"]: {bug_map[b["bug"]["id"]] for b in ad["bugs"]["bugs"]} for ad
-                      in advisories}
-        return result
```
thegreyd, line 213: @joepvd Earlier in this method we were doing a combination of async calls and sync calls depending on jira/bz. As you can see both branches went into the nitty gritty, doing individual ET call for jira issue / parsing bug ids from advisory. We now have both of those things supported by errata_tool library, having them use the same methods that we use everywhere else in the codebase is a win for maintainability - and also testing. Since right now to support this I would add json parsing of jira issues - which errata_tool object invocation already does that for us (I added that support :)). The only "regression" if you want to call it that, is it's going to be slow - stuff is not going to be async, so result will be slower by a few seconds - I think that's a fair tradeoff for now. I definitely want to look into how we can make errata_tool calls async - in a different PR though

elliottlib/cli/verify_attached_bugs_cli.py
```diff
@@ -180,41 +199,46 @@ async def _verify_attached_flaws_for(self, advisory_id: int, attached_trackers:
...
-
-    def filter_bugs_by_product(self, bugs):
-        # filter out bugs for different product (presumably security flaw bugs)
-        return [b for b in bugs if b.product == self.product]
+        green_print(f"Retrieving bugs for advisories: {advisory_ids}")
+        advisories = [Erratum(errata_id=advisory_id) for advisory_id in advisory_ids]
+
+        attached_bug_map = {advisory_id: set() for advisory_id in advisory_ids}
+        for bug_tracker_type in ['jira', 'bugzilla']:
```
thegreyd, line 232: Having common abstract methods for both bug trackers means we don't have to care about which one is which, as long we have the interface defined

elliottlib/cli/verify_attached_bugs_cli.py
```diff
@@ -243,13 +270,22 @@ def is_next_target(target_v):
             return pattern.match(target_v) and minor_version_tuple(target_v) == next_version
 
         # retrieve blockers and filter to those with correct product and target version
-        blockers = [b for b in self.bug_tracker.get_bugs(sorted(list(candidate_blockers))) if b.product == self.product]
```
thegreyd, line 246: replace b.product == self.product into is_ocp_bug()

elliottlib/cli/verify_attached_bugs_cli.py
vfreex: Yes, you are correct. Thanks for fixing this!

elliottlib/cli/verify_attached_bugs_cli.py
```diff
@@ -232,8 +262,11 @@ def filter_bugs_by_release(self, bugs: Iterable[Bug], complain: bool = False) ->
 
     def _get_blocking_bugs_for(self, bugs):
         # get blocker bugs in the next version for all bugs we are examining
-        candidate_blockers = [b.depends_on for b in bugs if b.depends_on]
-        candidate_blockers = {b for deps in candidate_blockers for b in deps}
+        candidate_blockers = []
+        for b in bugs:
+            if b.depends_on:
+                candidate_blockers.extend(b.depends_on)
```
vfreex, line 268: Any reason to reimplement this with a loop? it looks more verbose.

vfreex (commented):
Thanks for fixing this! Generally LGTM.

PR #415, opened 2022-09-06
Improve handling of CVE associations

joepvd:
This commit is an incomplete step towards ART-4552. `elliott
attach-cve-flaws` will now:
- error when mapping of builds to CVEs could not be made
- leave golang bumps to match the whole advisory

In case this runs into an error because the tracker is attached to the
wrong advisory, the tracker needs to be moved manually, and the process
repeated.

### DIFF
--- a/elliottlib/cli/attach_cve_flaws_cli.py
+++ b/elliottlib/cli/attach_cve_flaws_cli.py
@@ -123,17 +123,14 @@ async def get_flaws(runtime, advisory, bug_tracker, flaw_bug_tracker, noop):
     runtime.logger.info('Associating CVEs with builds')
     errata_config = runtime.get_errata_config()
     errata_api = AsyncErrataAPI(errata_config.get("server", constants.errata_url))
-    association_errors = []
     try:
         await errata_api.login()
         await associate_builds_with_cves(errata_api, advisory, attached_tracker_bugs, tracker_flaws, flaw_id_bugs, noop)
     except ValueError as e:
         runtime.logger.warn(f"Error associating builds with cves: {e}")
-        association_errors.append(e)
+        raise ValueError(f'Error associating builds with CVEs: {e}')
     finally:
         await errata_api.close()
-    if association_errors:
-        raise ValueError(f'Error associating builds with CVEs: {association_errors}')
     return first_fix_flaw_bugs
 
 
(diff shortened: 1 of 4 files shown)


elliottlib/cli/attach_cve_flaws_cli.py
```diff
@@ -123,13 +123,17 @@ async def get_flaws(runtime, advisory, bug_tracker, flaw_bug_tracker, noop):
...
     try:
         await errata_api.login()
         await associate_builds_with_cves(errata_api, advisory, attached_tracker_bugs, tracker_flaws, flaw_id_bugs, noop)
     except ValueError as e:
         runtime.logger.warn(f"Error associating builds with cves: {e}")
+        association_errors.append(e)
     finally:
         await errata_api.close()
+    if association_errors:
+        raise ValueError(f'Error associating builds with CVEs: {association_errors}')
```
sosiouxme, line 136: this makes it look like you'll end up with an array of these, when you'll only ever get one. why not just:
```suggestion
    try:
        await errata_api.login()
        await associate_builds_with_cves(errata_api, advisory, attached_tracker_bugs, tracker_flaws, flaw_id_bugs, noop)
    except ValueError as e:
        runtime.logger.warn(f"Error associating builds with cves: {e}")
        raise ValueError(f'Error associating builds with CVEs: {e}')
    finally:
        await errata_api.close()
```

elliottlib/errata_async.py
```diff
@@ -109,7 +109,7 @@ def compute_cve_package_exclusions(cls, attached_builds: Iterable[str], expected
         :return: a dict that key is CVE name, value is another dict with package name as key and 0 as value
         """
         brew_components = {parse_nvr(nvr)["name"] for nvr in attached_builds}
-        missing_brew_brew_components = {c for components in expected_cve_components.values() for c in components} - brew_components
+        missing_brew_brew_components = {c for components in expected_cve_components.values() for c in components if c != 'openshift-golang-builder-container'} - brew_components
```
sosiouxme, line 112: I don't like seeing specific names of things in code like this (though I'm guilty too). How about pulling this out to a constant, where we can supply a helpful comment about why it's there...

sosiouxme (commented):
some nits but in principle seems good

PR #416, opened 2022-09-07
fix sub_component not loaded

Ximinhan:
https://github.com/python-bugzilla/python-bugzilla/blob/main/bugzilla/bug.py#L87
https://coreos.slack.com/archives/GDBRP5YJH/p1662557859518319?thread_ts=1662540699.647459&cid=GDBRP5YJH


thegreyd:
Superceded by https://github.com/openshift/elliott/pull/417

PR #423, opened 2022-09-14
[ART-4554] use errata_tool lib

Ximinhan:


elliottlib/errata.py
```diff
@@ -635,25 +635,17 @@ def add_jira_bugs_with_retry(advisory_id: int, bugids: List[str], noop: bool = F
         if noop:
             logger.info('Dry run: Would have attached bugs')
             continue
-        results = add_multi_jira_issues(advisory_id, chunk_of_bugs)
-        for i, result in enumerate(results):
-            if result.status_code != 201:
-                rt = add_jira_issue(advisory_id, chunk_of_bugs[i])
-                if rt.status_code != 201:
-                    errata_error = rt.json()
```
thegreyd, line 643: ximin, I'd like to keep the error checking logic we have here

Ximinhan: I think errata_tool already handled some of the check https://github.com/red-hat-storage/errata-tool/blob/master/errata_tool/connector.py#L201

Ximinhan: as for the issue of bug already in another advisory, that is the bug of we didn't correctly handle placeholder jira bug, and we now excluded it in bug search, that situation may not happened again

elliottlib/bzutil.py
```diff
@@ -606,8 +607,11 @@ def remove_bugs(self, advisory_obj, bugids: List, noop=False):
         if noop:
             print(f"Would've removed bugs: {bugids}")
             return
-        advisory_id = advisory_obj.errata_id
-        return errata.remove_multi_jira_issues(advisory_id, bugids)
+        try:
+            advisory_obj.removeJIRAIssues(bugids)
+            advisory_obj.commit()
+        except ErrataException as e:
+            raise exceptions.ElliottFatalError(getattr(e, 'message', repr(e)))
```
vfreex, line 614: I didn't see a strong reason to capture and reraise a different exception here. This will lose the stack trace info.

elliottlib/errata.py
```diff
@@ -635,25 +635,21 @@ def add_jira_bugs_with_retry(advisory_id: int, bugids: List[str], noop: bool = F
...
-                        if 'The issue is filed already in' not in ' '.join(err):
-                            raise exceptions.ElliottFatalError(f"attach jira bug {chunk_of_bugs[i]} failed with "
-                                                               f"status={rt.status_code} "
-                                                               f"errmsg={rt.json()}")
+        try:
+            advisory.addJiraIssues(chunk_of_bugs)
+            advisory.commit()
+        except ErrataException as e:
+            logger.info(f"ErrataException Message: {e}\nRetrying...")
+            try:
```
vfreex, line 643: The unconditional retry logic here doesn't smell good. Would like to keep the original approach of handling exceptions or just let it crash if something wrong happens.

vfreex (changes requested):
There are a few concerns about handling exceptions, otherwise lgtm.

Ximinhan:
The purpose of using errata_tool is to handle the complexity of the API, including exception handling, so if the advisory object commit fails, let it happen

elliottlib/bzutil.py
```diff
@@ -606,8 +607,11 @@ def remove_bugs(self, advisory_obj, bugids: List, noop=False):
         if noop:
             print(f"Would've removed bugs: {bugids}")
             return
-        advisory_id = advisory_obj.errata_id
-        return errata.remove_multi_jira_issues(advisory_id, bugids)
+        try:
+            advisory_obj.removeJIRAIssues(bugids)
+            advisory_obj.commit()
+        except ErrataException as e:
+            raise e
```
joepvd, line 614: Why have this in a `try` block if the `except` just reraises the exception?

The bugzilla code still does a `return`. Any reason to drop that here?

elliottlib/errata.py
```diff
@@ -635,25 +635,11 @@ def add_jira_bugs_with_retry(advisory_id: int, bugids: List[str], noop: bool = F
...
-                    if 'idsfixed' not in errata_error['errors'] or len(errata_error['errors']) != 1:
-                        raise exceptions.ElliottFatalError(f"attach jira bug {chunk_of_bugs[i]} failed with "
-                                                           f"status={rt.status_code} "
-                                                           f"errmsg={rt.json()}")
-                    for err in errata_error['errors']['idsfixed']:
-                        if 'The issue is filed already in' not in ' '.join(err):
-                            raise exceptions.ElliottFatalError(f"attach jira bug {chunk_of_bugs[i]} failed with "
-                                                               f"status={rt.status_code} "
-                                                               f"errmsg={rt.json()}")
+        try:
```
joepvd, line 638: Looks like an unnecessary `try` block here.

joepvd: How will this code react to a jira issue that is already attached to a not shipped advisory?

joepvd (changes requested):
Some things to still look at

elliottlib/bzutil.py
Ximinhan: I don't think we need `return` here, if no exception raised it means success

joepvd: Think there's value in having the function signature equal between the two bugzilla and jira implementations.

elliottlib/errata.py
Ximinhan: removed try block, it will raise an exception, and I think we will not meet that issue https://github.com/openshift/elliott/pull/423/files#r971642699

joepvd: You might be right, but these status codes do not tell me much. You able to link to how these status codes are produced upstream? https://errata.devel.redhat.com/developer-guide/api-http-api.html is not revealing.

elliottlib/bzutil.py
Ximinhan: do you mean the function `remove_bugs` for bugzilla tracker? https://github.com/openshift/elliott/blob/master/elliottlib/bzutil.py#L704 it's using post api to implement it so we need the response, errata_tool have the function `removeBugs` we can also use errata_tool for it https://github.com/red-hat-storage/errata-tool/blob/master/errata_tool/erratum.py#L616

elliottlib/errata.py
Ximinhan: I guess this is it https://errata.devel.redhat.com/developer-guide/api-http-api.html#api-error-handling
in general, errata_tool is a wrapper of errata_api, The way we used to do it through the original API should now be simplified through errata_tool

elliottlib/errata.py
```diff
@@ -635,25 +635,8 @@ def add_jira_bugs_with_retry(advisory_id: int, bugids: List[str], noop: bool = F
...
-                    if 'idsfixed' not in errata_error['errors'] or len(errata_error['errors']) != 1:
-                        raise exceptions.ElliottFatalError(f"attach jira bug {chunk_of_bugs[i]} failed with "
-                                                           f"status={rt.status_code} "
-                                                           f"errmsg={rt.json()}")
-                    for err in errata_error['errors']['idsfixed']:
-                        if 'The issue is filed already in' not in ' '.join(err):
-                            raise exceptions.ElliottFatalError(f"attach jira bug {chunk_of_bugs[i]} failed with "
-                                                               f"status={rt.status_code} "
-                                                               f"errmsg={rt.json()}")
+        advisory.addJiraIssues(chunk_of_bugs)
```
vfreex, line 638: I think there is one thing missing here.
`chunk_of_bugs` may contain bugs that have been attached to the advisory. We need to ensure
- `advisory.addJiraIssues` doesn't panic if chunk_of_bugs contains already attached bugs.
- Or, ensure that chunk_of_bugs doesn't contain attached bugs.

vfreex (changes requested):

vfreex (approved):
/lgtm

PR #425, opened 2022-09-15
[ART-3778] sort CVEs in description

Ximinhan:
sort CVE bug list by it's impact and CVE number

### DIFF
--- a/elliottlib/bzutil.py
+++ b/elliottlib/bzutil.py
@@ -1058,6 +1058,14 @@ def get_highest_security_impact(bugs):
     return 'Low'
 
 
+def sort_cve_bugs(bugs):
+    def cve_sort_key(bug):
+        impact = constants.security_impact_map[get_highest_security_impact([bug])]
+        year, num = bug.alias[0].split("-")[1:]
+        return impact, -int(year), int(num)
+    return sorted(bugs, key=cve_sort_key, reverse=True)
+
+
 def is_first_fix_any(bugtracker, flaw_bug, current_target_release):
     """
     Check if a flaw bug is considered a first-fix for a GA target release
--- a/tests/test_bzutil.py
+++ b/tests/test_bzutil.py
@@ -421,6 +421,25 @@ class TestBZUtil(unittest.TestCase):
         actual = bzutil.is_first_fix_any(bzapi, bug, tr)
         self.assertEqual(expected, actual)
 
+    def test_sort_cve_bugs(self):
+        flaw_bugs = [
+            flexmock(alias=['CVE-2022-123'], severity='midium'),
+            flexmock(alias=['CVE-2022-456'], severity='urgent'),
+            flexmock(alias=['CVE-2021-789'], severity='midium'),
+            flexmock(alias=['CVE-2021-100'], severity='midium')
+        ]
+        sort_list = bzutil.sort_cve_bugs(flaw_bugs)
+        expected = [
+            flexmock(alias=['CVE-2022-456'], severity='urgent'),
+            flexmock(alias=['CVE-2021-789'], severity='midium'),
+            flexmock(alias=['CVE-2021-100'], severity='midium'),
+            flexmock(alias=['CVE-2022-123'], severity='midium')
+        ]
+        self.assertEqual(expected[0].alias, sort_list[0].alias)
+        self.assertEqual(expected[1].alias, sort_list[1].alias)
+        self.assertEqual(expected[2].alias, sort_list[2].alias)
+        self.assertEqual(expected[3].alias, sort_list[3].alias)
+
     def test_is_first_fix_any_no_valid_trackers(self):
         tr = '4.8.0'
         tracker_bug_ids = [1, 2]
--- a/tests/test_bzutil.py
+++ b/tests/test_bzutil.py
@@ -423,17 +423,17 @@ class TestBZUtil(unittest.TestCase):
 
     def test_sort_cve_bugs(self):
         flaw_bugs = [
-            flexmock(alias=['CVE-2022-123'], severity='midium'),
+            flexmock(alias=['CVE-2022-123'], severity='Low'),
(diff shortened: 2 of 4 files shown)


thegreyd:
Nice! Can you add some basic unit tests @Ximinhan ?

elliottlib/bzutil.py
```diff
@@ -1058,6 +1058,29 @@ def get_highest_security_impact(bugs):
...
+    """
+    buga_l = constants.security_impact_map[get_highest_security_impact([bug_a])]
+    bugb_l = constants.security_impact_map[get_highest_security_impact([bug_b])]
+    if buga_l == bugb_l:
+        return int(bug_a.alias[0].replace("CVE", "").replace("-", "")) > int(bug_b.alias[0].replace("CVE", "").replace("-", ""))
+    else:
+        return buga_l < bugb_l
+
+
+def sort_cve_bugs(bugs):
```
vfreex, line 1075: Consider implementing with builtin sort functions instead of homebrewed bubble sort. https://docs.python.org/3/howto/sorting.html

elliottlib/bzutil.py
```diff
@@ -1058,6 +1058,29 @@ def get_highest_security_impact(bugs):
...
+def compare_impact(bug_a, bug_b):
+    """
+    Check if CVE priority of bug_a if lower than bug_b
+    first compare severity level, if severity level is the same
+    then compare CVE number
+    """
+    buga_l = constants.security_impact_map[get_highest_security_impact([bug_a])]
+    bugb_l = constants.security_impact_map[get_highest_security_impact([bug_b])]
+    if buga_l == bugb_l:
+        return int(bug_a.alias[0].replace("CVE", "").replace("-", "")) > int(bug_b.alias[0].replace("CVE", "").replace("-", ""))
```
vfreex, line 1070: This doesn't seem to be correct. Consider `CVE-2022-1234` and `CVE-2011-12345`, the result will be wrong.
Needs to compare per digit rather than concatenate all digits as a whole.

vfreex (changes requested):

elliottlib/bzutil.py
Ximinhan: updated

elliottlib/bzutil.py
```diff
@@ -1058,6 +1059,27 @@ def get_highest_security_impact(bugs):
...
+    buga_l = constants.security_impact_map[get_highest_security_impact([bug_a])]
+    bugb_l = constants.security_impact_map[get_highest_security_impact([bug_b])]
+    if buga_l == bugb_l:
+        if bug_a.alias[0].split("-")[1] == bug_b.alias[0].split("-")[1]:
+            return 1 if int(bug_a.alias[0].split("-")[2]) < int(bug_b.alias[0].split("-")[2]) else -1
+        else:
+            return 1 if int(bug_a.alias[0].split("-")[1]) < int(bug_b.alias[0].split("-")[1]) else -1
+    else:
+        return 1 if buga_l > bugb_l else -1
+
```
vfreex, line 1077: This can be simplified further.

Python supports comparing two lists or tuples:
```python
(2021, 1) < (2021, 2)
(2022, 1) > (2021, 2)
```

So something like the following should work:

```suggestion
def cve_sort_key(bug):
    impact = constants.security_impact_map[get_highest_security_impact([bug])]
    year, num = bug.alias[0].split("-")[1:]
    return impact, year, num
```

Then sort it with:
return sorted(bugs, key=cve_sort_key, reverse=True)

Ximinhan: thanks for you suggestion, I updated

tests/test_bzutil.py
```diff
@@ -421,6 +421,25 @@ def test_is_first_fix_any_validate(self):
...
 
+    def test_sort_cve_bugs(self):
+        flaw_bugs = [
+            flexmock(alias=['CVE-2022-123'], severity='midium'),
+            flexmock(alias=['CVE-2022-456'], severity='urgent'),
+            flexmock(alias=['CVE-2021-789'], severity='midium'),
+            flexmock(alias=['CVE-2021-100'], severity='midium')
+        ]
+        sort_list = bzutil.sort_cve_bugs(flaw_bugs)
+        expected = [
```
joepvd, line 432: There's a typo: `midium` is not expected. I think `Low` is an impact, are you sure these words are capitalized correctly?
Sorting should happen first on Severity, and then on Alias alphabetically. Think the test does not reflect that.

elliottlib/bzutil.py
```diff
@@ -1058,6 +1058,14 @@ def get_highest_security_impact(bugs):
     return 'Low'
 
 
+def sort_cve_bugs(bugs):
+    def cve_sort_key(bug):
+        impact = constants.security_impact_map[get_highest_security_impact([bug])]
+        year, num = bug.alias[0].split("-")[1:]
+        return impact, -int(year), int(num)
+    return sorted(bugs, key=cve_sort_key, reverse=True)
```
joepvd, line 1066: How are `Low` vs `Critical` vs `Urgent` impacts sorted here?

joepvd (changes requested):
Think some tweaks are still needed.

elliottlib/bzutil.py
Ximinhan: they will be translated to number by https://github.com/Ximinhan/elliott/blob/a898f60bd397ddf8130a6e2e658bb2c0ffd01225/elliottlib/constants.py#L33
so Critical and Urgent have the same value

tests/test_bzutil.py
Ximinhan: severity have different value in bugzilla and jira, and we use function `constants.security_impact_map[get_highest_security_impact([bug])] `o translate it to integer so they can compare, this test reflect the severity value was sorted and cve number also sorted

vfreex: Is it `medium`?

vfreex (approved):

PR #428, opened 2022-09-15
Fix issues introduced by #426

vfreex:
1. AsyncErrataAPI shouldn't be closed inside of the loop because it is created outside of the loop. Use new AsyncErrataAPI instance for each advisory.
2. Don't panic on advisories without any trackers.


thiagoalessio (approved):
/lgtm

PR #434, opened 2022-09-22
[ART-4554] use errata_tool add bugs

Ximinhan:
filter out attached bugs then add bugs with errata_tools
wait until we got this https://github.com/red-hat-storage/errata-tool/pull/234

### DIFF
--- a/elliottlib/bzutil.py
+++ b/elliottlib/bzutil.py
@@ -22,6 +22,8 @@ from elliottlib import constants, exceptions, exectools, logutil, errata, util
 from elliottlib.cli import cli_opts
 from elliottlib.metadata import Metadata
 from elliottlib.util import isolate_timestamp_in_release
+from errata_tool.jira_issue import JiraIssue as ErrataJira
+from errata_tool.bug import Bug as ErrataBug
 
 logger = logutil.getLogger(__name__)
 
@@ -427,6 +429,9 @@ class BugTracker:
     def get_flaw_bugs(self, bug_ids: List, strict: bool = True, verbose: bool = False):
         raise NotImplementedError
 
+    def filter_attached_bugs(self, bug_ids: List):
+        raise NotImplementedError
+
 
 class JIRABugTracker(BugTracker):
     @staticmethod
@@ -592,8 +597,8 @@ class JIRABugTracker(BugTracker):
         if noop:
             print(f"Would've removed bugs: {bugids}")
             return
-        advisory_id = advisory_obj.errata_id
-        return errata.remove_multi_jira_issues(advisory_id, bugids)
+        advisory_obj.removeJIRAIssues(bugids)
+        advisory_obj.commit()
 
     def attach_bugs(self, advisory_id: int, bugids: List, noop=False, verbose=False):
         return errata.add_jira_bugs_with_retry(advisory_id, bugids, noop=noop)
@@ -620,6 +625,9 @@ class JIRABugTracker(BugTracker):
 
     def get_flaw_bugs(self, bug_ids: List, strict: bool = True, verbose: bool = False):
         return [b for b in self.get_bugs(bug_ids, permissive=not strict, verbose=verbose) if b.is_flaw_bug()]
+
+    def filter_attached_bugs(self, bug_ids: List):
+        return [b for b in bug_ids if not ErrataJira(b).all_advisory_ids]
 
 
 class BugzillaBugTracker(BugTracker):
@@ -833,6 +841,9 @@ class BugzillaBugTracker(BugTracker):
         fields = ["product", "component", "depends_on", "alias", "severity", "summary"]
         return [b for b in self.get_bugs(bug_ids, permissive=not strict, include_fields=fields, verbose=verbose) if
                 b.is_flaw_bug()]
+
+    def filter_attached_bugs(self, bug_ids: List):
+        return [b for b in bug_ids if not ErrataBug(b).all_advisory_ids]
 
 
 def get_highest_impact(trackers, tracker_flaws_map):
--- a/elliottlib/errata.py
+++ b/elliottlib/errata.py
@@ -578,6 +578,7 @@ def add_bugzilla_bugs_with_retry(advisory_id: int, bugids: List, noop: bool = Fa
 
     existing_bugs = bzutil.BugzillaBugTracker.advisory_bug_ids(advisory)
     new_bugs = set(bugids) - set(existing_bugs)
+    new_bugs = bzutil.BugzillaBugTracker.filter_attached_bugs(new_bugs)
     logger.info(f'New bugs (not already attached to advisory): {len(new_bugs)}')
     logger.debug(f'New bugs: {sorted(new_bugs)}')
     logger.debug(f'Bugs already attached: {sorted(set(bugids) & set(existing_bugs))}')
@@ -626,6 +627,7 @@ def add_jira_bugs_with_retry(advisory_id: int, bugids: List[str], noop: bool = F
 
     existing_bugs = bzutil.JIRABugTracker.advisory_bug_ids(advisory)
     new_bugs = set(bugids) - set(existing_bugs)
+    new_bugs = bzutil.JIRABugTracker.filter_attached_bugs(new_bugs)
     logger.info(f'New bugs (not already attached to advisory): {len(new_bugs)}')
     logger.debug(f'New bugs: {sorted(new_bugs)}')
     logger.debug(f'Bugs already attached: {sorted(set(bugids) & set(existing_bugs))}')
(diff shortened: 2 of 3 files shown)


elliottlib/bzutil.py
```diff
@@ -635,6 +660,9 @@ def get_tracker_bugs(self, bug_ids: List, strict: bool = False, verbose: bool =
     def get_flaw_bugs(self, bug_ids: List, strict: bool = True, verbose: bool = False):
         return [b for b in self.get_bugs(bug_ids, permissive=not strict, verbose=verbose) if b.is_flaw_bug()]
 
+    def filter_attached_bugs(self, bug_ids: List, strict: bool = True, verbose: bool = False):
+        return [b.id for b in self.get_bugs(bug_ids, permissive=not strict, verbose=verbose) if not b.is_attached()]
```
joepvd, line 664: Did you try this? I think `@property` means that the behavior is not called as a method with `()`, but as a property.

Ximinhan: updated

joepvd (changes requested):
Let's use the errata end points to determine if a bug is attached to an advisory:

```console
$ curl -sSL --negotiate --user : https://errata.devel.redhat.com/jira_issues/OCPBUGS-1590/advisories.json | jq -r '.[].advisory_name'
RHBA-2022:6661

$ curl -sSL --negotiate --user : https://errata.devel.redhat.com/bugs/2088196/advisories.json | jq -r '.[].advisory_name'
RHSA-2022:2272
```

Ximinhan:
upstream pr merged, updated in this pr to use latest errata_tool

elliottlib/errata.py
```diff
@@ -626,6 +628,8 @@ def add_jira_bugs_with_retry(advisory_id: int, bugids: List[str], noop: bool = F
 
     existing_bugs = bzutil.JIRABugTracker.advisory_bug_ids(advisory)
     new_bugs = set(bugids) - set(existing_bugs)
+    logger.info(f'New bugs before filer out already attached: {sorted(new_bugs)}')
```
joepvd, line 631: typo

elliottlib/errata.py
```diff
@@ -578,6 +578,8 @@ def add_bugzilla_bugs_with_retry(advisory_id: int, bugids: List, noop: bool = Fa
 
     existing_bugs = bzutil.BugzillaBugTracker.advisory_bug_ids(advisory)
     new_bugs = set(bugids) - set(existing_bugs)
+    logger.info(f'New bugs before filer out already attached: {sorted(new_bugs)}')
```
joepvd, line 581: typo

elliottlib/bzutil.py
```diff
@@ -22,6 +22,8 @@
 from elliottlib.cli import cli_opts
 from elliottlib.metadata import Metadata
 from elliottlib.util import isolate_timestamp_in_release
+from errata_tool.jira_issue import JiraIssue
```
joepvd, line 25: Maybe give this another name that tells that it comes from Errata world. `ErrataJira` maybe?

elliottlib/bzutil.py
```diff
@@ -22,6 +22,8 @@
...
+from errata_tool.bug import Bug as Errata_Bug
```
joepvd, line 26: Keep class names CamelCase

elliottlib/bzutil.py
```diff
@@ -621,6 +626,9 @@ def get_tracker_bugs(self, bug_ids: List, strict: bool = False, verbose: bool =
     def get_flaw_bugs(self, bug_ids: List, strict: bool = True, verbose: bool = False):
         return [b for b in self.get_bugs(bug_ids, permissive=not strict, verbose=verbose) if b.is_flaw_bug()]
 
+    def filter_attached_bugs(self, bug_ids: List, strict: bool = True, verbose: bool = False):
+        return [b for b in bug_ids if JiraIssue(b).all_advisory_ids != []]
```
joepvd, line 630: Think this is more pythonic than testing equalisty with empty list: `if not JiraIssue(b).all_advisory_ids`

joepvd (commented):
Let's try to get this in before this week's releases. Have some nits, still.
Did you manage to test this out?

Ximinhan:
I tested the `errata_tool.jira_issue` lib and `errata_tool.bug` lib with errata_tool version 1.29.0

elliottlib/bzutil.py
```diff
@@ -621,6 +626,9 @@ def get_tracker_bugs(self, bug_ids: List, strict: bool = False, verbose: bool =
     def get_flaw_bugs(self, bug_ids: List, strict: bool = True, verbose: bool = False):
         return [b for b in self.get_bugs(bug_ids, permissive=not strict, verbose=verbose) if b.is_flaw_bug()]
 
+    def filter_attached_bugs(self, bug_ids: List, strict: bool = True, verbose: bool = False):
```
thegreyd, line 629: I'm thinking this would be better off as a static method in JIRABug and BugzillaBug , since it's not interacting with the BugTracker functionality?

elliottlib/bzutil.py
```diff
@@ -834,6 +842,9 @@ def get_flaw_bugs(self, bug_ids: List, strict: bool = True, verbose: bool = Fals
         return [b for b in self.get_bugs(bug_ids, permissive=not strict, include_fields=fields, verbose=verbose) if
                 b.is_flaw_bug()]
 
+    def filter_attached_bugs(self, bug_ids: List, strict: bool = True, verbose: bool = False):
+        return [b for b in bug_ids if ErrataBug(b).all_advisory_ids != []]
```
thegreyd, line 846: Shouldn't this be `if ErrataBug(b).all_advisory_ids == []`?

elliottlib/errata.py
```diff
@@ -578,6 +578,8 @@ def add_bugzilla_bugs_with_retry(advisory_id: int, bugids: List, noop: bool = Fa
 
     existing_bugs = bzutil.BugzillaBugTracker.advisory_bug_ids(advisory)
     new_bugs = set(bugids) - set(existing_bugs)
+    logger.info(f'New bugs before filter out already attached: {sorted(new_bugs)}')
```
thegreyd, line 581: I don't see a reason to print out bugs before filtering

elliottlib/errata.py
```diff
@@ -626,6 +628,8 @@ def add_jira_bugs_with_retry(advisory_id: int, bugids: List[str], noop: bool = F
 
     existing_bugs = bzutil.JIRABugTracker.advisory_bug_ids(advisory)
     new_bugs = set(bugids) - set(existing_bugs)
+    logger.info(f'New bugs before filter out already attached: {sorted(new_bugs)}')
```
thegreyd, line 631: IMO we can skip this message, doesn't provide much value

thegreyd:
Thanks for the PR ximin. But I'd like us to evaluate if this is better - multiple api calls (for each bug so could be hundreds) to ET is better than getting an batch error message from ET? I don't know.. could be an issue at GA time. Thoughts? Have you seen any performance issues?

elliottlib/bzutil.py
Ximinhan: This method needs to distinguish between different bug types, I think it is appropriate to put them in different bugtrackers, if move the method into the bug class then it will instantiate each bug object while this will only need a bug id.

Ximinhan:
good point, for each bug it will cost around 2 seconds to check its advisory_ids, so for regular release, it may cost 1 min but at GA time it may cost a lot of time, an option is don't filter the bugs if the bug list is too long( like over 200 )

Ximinhan:
I will try if async can be used here

vfreex (approved):
/lgtm

elliottlib/bzutil.py
```diff
@@ -834,6 +842,9 @@ def get_flaw_bugs(self, bug_ids: List, strict: bool = True, verbose: bool = Fals
         return [b for b in self.get_bugs(bug_ids, permissive=not strict, include_fields=fields, verbose=verbose) if
                 b.is_flaw_bug()]
 
+    def filter_attached_bugs(self, bug_ids: List, strict: bool = True, verbose: bool = False):
```
vfreex, line 845: `verbose` is not honored. Can be removed.

vfreex (changes requested):
1 minor issue. otherwise lgtm.

vfreex (approved):

Ximinhan:
> Thanks for the PR ximin. But I'd like us to evaluate if this is better - multiple api calls (for each bug so could be hundreds) to ET is better than getting an batch error message from ET? I don't know.. could be an issue at GA time. Thoughts? Have you seen any performance issues?

I tested the performance, seems the constraint is on the server side, use async or not will both take around 2 seconds per bug to complete the request, at GA time if over a thousand bugs need to check it will take around half an hour to an hour to complete the check, but I think attaching bugs operation will take longer than this

joepvd:
Think this is mostly for the situation that a 4.Y.(Z-1) advisory has not yet shipped, and the bugs are taken into consideration. Maybe we can skip this process if it is a GA advisory? It might be that a bug is attached to another advisory that is not ART-managed, but that is good to run into. The fix is to exclude component from our queries.
The question then becomes: How can we detect if an assembly or an advisory is a GA advisory? Maybe by looking if `hyperkube` or `cluster-service-version-container` is in the shipped brew tag? A new assembly type? Or maybe something smarter?

thegreyd:
We can just look for the number of bugs exceeding a high enough threshold and decide to not use it

PR #437, opened 2022-09-23
raise error if CVE bug missing labels

Ximinhan:
if a bug missing tracker label bug have CVE-XXXX in summary, raise the suspicious bug

### DIFF
--- a/elliottlib/bzutil.py
+++ b/elliottlib/bzutil.py
@@ -58,6 +58,8 @@ class Bug:
         raise NotImplementedError
 
     def is_tracker_bug(self):
+        if not set(constants.TRACKER_BUG_KEYWORDS).issubset(set(self.keywords)) and re.search(r'^CVE-(\d+)-(\d+)', self.summary):
+            raise exceptions.ElliottFatalError(f"Bug {self.id} don't have tracker keywords while have CVE number in summary, need manual check again")
         return set(constants.TRACKER_BUG_KEYWORDS).issubset(set(self.keywords))
 
     def is_flaw_bug(self):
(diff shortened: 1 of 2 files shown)


joepvd (changes requested):
I think the test failures of previous iterations actually tell something. In this implementation, it is assumed that there is a summary, even if that field does not exist. This should either get initialized in `Bug().__init__`, or it should first be tested if the property exists.

Could you add a test for when this code path hits?

Also, could you rebase this into a single commit, and open the PR from your fork rather than from the openshift repository?

thiagoalessio (changes requested):
+1 in Joep's change requests.
Also, can we have more descriptive commit messages, and a corresponding ART JIRA (feel free to create one describing the bug in case it doesn't exist yet) ?

elliottlib/bzutil.py
```diff
@@ -58,6 +58,8 @@ def whiteboard_component(self):
         raise NotImplementedError
 
     def is_tracker_bug(self):
+        if not set(constants.TRACKER_BUG_KEYWORDS).issubset(set(self.keywords)) and re.search(r'^CVE-(\d+)-(\d+)', self.summary):
```
thegreyd, line 61: @Ximinhan Let's not have this logic in this method. I propose to have it in a separate function, 
```python
def looks_like_tracker_bug(self):
    return re.search(r'^CVE-(\d+)-(\d+)', self.summary)
```

thegreyd:
@Ximinhan We should think about where to raise this type of warning / issue. One place could be check-bugs job, and IMO this fits nicely there than say find-bugs:sweep. For that we would need to have this in verify-bugs and verify-attached-bugs command

PR #439, opened 2022-09-28
[ART-4877] Complain when Tracker Bugs don't have Tracker Keywords

Ximinhan:
ref: [ART-4877](https://issues.redhat.com//browse/ART-4877)
if a bug was backported to Jira, it might miss CVE-related labels but it has CVE feature in its summary and described in its description, so add a check to raise concern if we find such a mismatch in summary, it can be a filter that effect bug sweep and bug validate

### DIFF
--- a/elliottlib/bzutil.py
+++ b/elliottlib/bzutil.py
@@ -60,6 +60,9 @@ class Bug:
     def is_tracker_bug(self):
         return set(constants.TRACKER_BUG_KEYWORDS).issubset(set(self.keywords))
 
+    def is_cve_in_summary(self):
+        return re.search(r'^CVE-(\d+)-(\d+)', self.summary)
+
     def is_flaw_bug(self):
         return self.product == "Security Response" and self.component == "vulnerability"
 
--- a/elliottlib/bzutil.py
+++ b/elliottlib/bzutil.py
@@ -61,7 +61,7 @@ class Bug:
         return set(constants.TRACKER_BUG_KEYWORDS).issubset(set(self.keywords))
 
     def is_cve_in_summary(self):
-        return re.search(r'^CVE-(\d+)-(\d+)', self.summary)
+        return True if re.search(r'^CVE-\d+-\d+', self.summary) else False
 
     def is_flaw_bug(self):
         return self.product == "Security Response" and self.component == "vulnerability"
--- a/elliottlib/cli/find_bugs_sweep_cli.py
+++ b/elliottlib/cli/find_bugs_sweep_cli.py
@@ -260,8 +260,7 @@ def categorize_bugs_by_type(bugs: List[Bug], advisory_id_map: Dict[str, int], ma
 
     warning_bug = [b.id for b in non_tracker_bugs if b.is_cve_in_summary()]
     if warning_bug:
-        logger.warn(f"Bug {warning_bug} have CVE number in summary but don't have tracker keywords "
-                    "need to check if they are missing CVE labels")
+        logger.warn(f"Bug {warning_bug} has CVE number in summary but does not have tracker keywords")
 
     if not advisory_id_map:
         logger.info("Skipping sorting/attaching Tracker Bugs. Advisories with attached builds must be given to "
--- a/elliottlib/cli/find_bugs_sweep_cli.py
+++ b/elliottlib/cli/find_bugs_sweep_cli.py
@@ -258,6 +258,11 @@ def categorize_bugs_by_type(bugs: List[Bug], advisory_id_map: Dict[str, int], ma
     for b in tracker_bugs:
         logger.info((b.id, b.whiteboard_component))
 
+    warning_bug = [b.id for b in non_tracker_bugs if b.is_cve_in_summary()]
+    if warning_bug:
+        logger.warn(f"Bug {warning_bug} have CVE number in summary but don't have tracker keywords "
+                     "need to check if they are missing CVE labels")
+
     if not advisory_id_map:
         logger.info("Skipping sorting/attaching Tracker Bugs. Advisories with attached builds must be given to "
                     "validate trackers.")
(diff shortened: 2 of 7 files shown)


thegreyd:
Thanks for creating the story and this PR, taking a look today

Ximinhan:
ut test failed should be fixed by https://github.com/openshift/elliott/pull/441

tests/test_find_bugs_sweep_cli.py
```diff
@@ -210,11 +210,11 @@ class TestCategorizeBugsByType(unittest.TestCase):
     def test_categorize_bugs_by_type(self):
         advisory_id_map = {'image': 1, 'rpm': 2, 'extras': 3}
         bugs = [
-            flexmock(id='OCPBUGS-1', is_tracker_bug=lambda: True, whiteboard_component='foo'),
-            flexmock(id='OCPBUGS-2', is_tracker_bug=lambda: True, whiteboard_component='bar'),
-            flexmock(id='OCPBUGS-3', is_tracker_bug=lambda: True, whiteboard_component='buzz'),
-            flexmock(id='OCPBUGS-4', is_tracker_bug=lambda: False),
-            flexmock(id='OCPBUGS-5', is_tracker_bug=lambda: False)
+            flexmock(id='OCPBUGS-1', is_tracker_bug=lambda: True, is_cve_in_summary=lambda: True, whiteboard_component='foo'),
```
joepvd, line 213: Think this is a sign that the summary is not initialized properly. I think the property should return an empty string if that jira field is not set. Then it is not necessary to patch this method all over the place and distract from what is actually being tested.

tests/test_verify_attached_bugs_cli.py
```diff
@@ -59,7 +63,7 @@ def test_verify_bugs_with_sweep_cli(self):
             .with_args(bugs)\
             .and_return(blocking_bugs_map)
 
-        result = runner.invoke(cli, ['-g', 'openshift-4.6', '--assembly=stream', 'verify-bugs'])
+        result = runner.invoke(cli, ['-g', 'openshift-4.6', '--assembly=stream', '--debug', 'verify-bugs'])
```
joepvd, line 66: What is the motivation for the `--debug`?

tests/test_find_bugs_sweep_cli.py
Ximinhan: verify function is added to validate, so bugs require this method is appropriate

joepvd: I am not sure I follow. 
If the summary is properly initialized, there is no need to patch this function.
I think the option of initializing the summary in the class is better than needing to patch this function in any unit test that run that touch that function.

elliottlib/bzutil.py
```diff
@@ -60,6 +60,9 @@ def whiteboard_component(self):
     def is_tracker_bug(self):
         return set(constants.TRACKER_BUG_KEYWORDS).issubset(set(self.keywords))
 
+    def is_cve_in_summary(self):
+        return True if re.search(r'^CVE-\d+-\d+', self.summary) else False
```
thegreyd, line 64:
```suggestion
        return re.search(r'^CVE-\d+-\d+', self.summary)
```

elliottlib/cli/find_bugs_sweep_cli.py
```diff
@@ -258,6 +258,10 @@ def categorize_bugs_by_type(bugs: List[Bug], advisory_id_map: Dict[str, int], ma
     for b in tracker_bugs:
         logger.info((b.id, b.whiteboard_component))
 
+    warning_bug = [b.id for b in non_tracker_bugs if b.is_cve_in_summary()]
+    if warning_bug:
+        logger.warn(f"Bug {warning_bug} has CVE number in summary but dose not have tracker keywords")
```
thegreyd, line 263:
```suggestion
        logger.warn(f"Bug {warning_bug} looks like a CVE tracker but does not have tracker keywords!")
```

tests/test_find_bugs_sweep_cli.py
```diff
@@ -210,11 +210,11 @@ class TestCategorizeBugsByType(unittest.TestCase):
...
+            flexmock(id='OCPBUGS-2', is_tracker_bug=lambda: True, is_cve_in_summary=lambda: True, whiteboard_component='bar'),
+            flexmock(id='OCPBUGS-3', is_tracker_bug=lambda: True, is_cve_in_summary=lambda: True, whiteboard_component='buzz'),
+            flexmock(id='OCPBUGS-4', is_tracker_bug=lambda: False, is_cve_in_summary=lambda: False),
+            flexmock(id='OCPBUGS-5', is_tracker_bug=lambda: False, is_cve_in_summary=lambda: False)
```
thegreyd, line 217: Can we make them all `is_cve_in_summary=lambda: True`? Making 2 of them False, makes you think that this is being tested while it isn't..

elliottlib/cli/verify_attached_bugs_cli.py
```diff
@@ -370,6 +370,12 @@ def _verify_bug_status(self, bugs):
                 status = f"{bug.status}: {bug.resolution}"
             self._complain(f"Bug {bug.id} has status {status}")
 
+    def _verify_bug_summary(self, bugs):
```
thegreyd, line 373: Can we have unit tests for these?

elliottlib/bzutil.py
```diff
@@ -60,6 +60,9 @@ def whiteboard_component(self):
     def is_tracker_bug(self):
         return set(constants.TRACKER_BUG_KEYWORDS).issubset(set(self.keywords))
 
+    def is_cve_in_summary(self):
```
thegreyd, line 63: Can we have unit tests for this?

thegreyd:
@Ximinhan Request to include the jira story link, and the commands affected and how to test them in the PR description - this make PR reviews much easier!

elliottlib/bzutil.py
Ximinhan: `re.search` returns a re object, joep suggests we should make the return value a clear boolean

elliottlib/cli/verify_attached_bugs_cli.py
Ximinhan: it can be tested in test_verify_bugs_with_sweep_cli test case

elliottlib/bzutil.py
Ximinhan: added

elliottlib/bzutil.py
thegreyd: Ah that's true then `return bool(re.search(r'^CVE-\d+-\d+', self.summary))` should do the job right?

Ximinhan: create a new pr https://github.com/openshift/elliott/pull/454

PR #440, opened 2022-09-28
Do not trigger on Proposed Blockers

joepvd:
Following the
[announcement](https://groups.google.com/a/redhat.com/g/aos-devel/c/0j22YV-18no)  that bugs of high enough priority/severity, Release Blocker gets set to Proposed automatically. We should stop alerting on this.


thegreyd:
Test fix being discussed at https://github.com/openshift/elliott/pull/441

PR #443, opened 2022-10-05
Sort CVEs better

joepvd:
This should follow the spec: Order by Impact, then order by int.

### DIFF
--- a/elliottlib/bzutil.py
+++ b/elliottlib/bzutil.py
@@ -1048,7 +1048,7 @@ def sort_cve_bugs(bugs):
     def cve_sort_key(bug):
         impact = constants.security_impact_map[get_highest_security_impact([bug])]
         year, num = bug.alias[0].split("-")[1:]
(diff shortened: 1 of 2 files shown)


elliottlib/bzutil.py
```diff
@@ -1046,10 +1046,10 @@ def get_highest_security_impact(bugs):
 
 def sort_cve_bugs(bugs):
     def cve_sort_key(bug):
-        impact = constants.security_impact_map[get_highest_security_impact([bug])]
-        year, num = bug.alias[0].split("-")[1:]
-        return impact, -int(year), int(num)
```
sosiouxme, line 1051: Think we could have just changed this, to get it to sort in the right order?
```
        return impact, -int(year), -int(num)
```

sosiouxme (changes requested):

sosiouxme (approved):
/lgtm

PR #446, opened 2022-10-12
Fix filter attached bugs

vfreex:


vfreex:
This is no longer needed because the PR that introduces this issue has been reverted.

PR #449, opened 2022-10-13
[RHELWF-7726] increase page size to 1000

Ximinhan:
elliott verify-attached-bugs will raise such an error message
```
On advisory 103179, CVE-2022-30322 is associated with Brew components openshift-enterprise-cli-alt-container, ose-insights-operator-container without a tracker bug. You may need to explicitly exclude those Brew components from the CVE mapping or attach the corresponding tracker bugs.
On advisory 103179, CVE-2022-30323 is associated with Brew components kuryr-controller-container, ose-azure-disk-csi-driver-container without a tracker bug. You may need to explicitly exclude those Brew components from the CVE mapping or attach the corresponding tracker bugs.
On advisory 103179, CVE-2022-26945 is associated with Brew components cluster-monitoring-operator-container, ose-network-metrics-daemon-container without a tracker bug. You may need to explicitly exclude those Brew components from the CVE mapping or attach the corresponding tracker bugs.
Some bug problems were listed above. Please investigate.
```
it looks like pagination is not correctly working as expected, the [mapping info](https://errata.devel.redhat.com/api/v1/cve_package_exclusion?filter[errata_id]=103179&page[size]=300&page[number]=2) is on page 2
the issue is pagination on the server is not correctly ordered, ticket to this bug is [RHELWF-7726](https://issues.redhat.com/browse/RHELWF-7726)

this PR increases the page size to the maximum value of 1000 to reduce the odds of this error

### DIFF
--- a/elliottlib/errata_async.py
+++ b/elliottlib/errata_async.py
@@ -62,7 +62,7 @@ class AsyncErrataAPI:
     async def get_cve_package_exclusions(self, advisory_id: int):
         path = "/api/v1/cve_package_exclusion"
         # This is a paginated API, we need to increment page[number] until an empty array is returned.
-        params = {"filter[errata_id]": str(int(advisory_id)), "page[number]": 1, "page[size]": 300}
+        params = {"filter[errata_id]": str(int(advisory_id)), "page[number]": 1, "page[size]": 1000}
         while True:
             result = await self._make_request(aiohttp.hdrs.METH_GET, path, params=params)
             data: List[Dict] = result.get('data', [])
--- a/tests/test_errata_async.py
+++ b/tests/test_errata_async.py
@@ -122,7 +122,7 @@ class TestAsyncErrataAPI(TestCase):
             return items
 
         actual = get_event_loop().run_until_complete(_call())
-        _make_request.assert_awaited_with(ANY, 'GET', '/api/v1/cve_package_exclusion', params={'filter[errata_id]': '1', 'page[number]': 3, 'page[size]': 300})
+        _make_request.assert_awaited_with(ANY, 'GET', '/api/v1/cve_package_exclusion', params={'filter[errata_id]': '1', 'page[number]': 3, 'page[size]': 1000})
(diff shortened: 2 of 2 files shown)


vfreex (commented):
I did more tests. I didn't see any wrong wrong with Elliott `get_cve_package_exclusions` pagination:

```python

import asyncio

from elliottlib import constants
from elliottlib.errata_async import AsyncErrataAPI


async def main():
    api = AsyncErrataAPI(constants.errata_url)
    await api.login()
    items = []
    async for item in api.get_cve_package_exclusions(103179):
        items.append(item)
    print(f"{len(items)}")


asyncio.run(main())
```

The problem is probably in Errata Tool server. Will have a closer look.

/hold

vfreex:
I created https://issues.redhat.com/browse/RHELWF-7726. 
As a workaround for this specific advisory, we can increase the page size to a value like 600 (or some large value, needs tests) to avoid pagination.

Ximinhan:
Increasing the size can't avoid the issue, we should get all content in one request to avoid pagination

Ximinhan:
Increasing the size can't avoid the issue, we should get all content in one request to avoid pagination

Ximinhan:
I don't want to wait for a long time for the server side to solve it. We can solve our current issue through this pr

vfreex (changes requested):
> Increasing the size can't avoid the issue, we should get all content in one request to avoid pagination

Can you justify this statement with example API calls? I didn't see any issues by increasing page size.

> I don't want to wait for a long time for the server side to solve it. We can solve our current issue through this pr.

I don't think so. This PR fetches the first page then refetches with page size of result count it if it paginated. It will panic if page size is too large.

Ximinhan:
all right, I didn't mean the current approach is wrong, but almost every time I saw this issue if we need the second page
we can increase the page size to the maximum value of 1000 if the result count is over 1000 we will see this again

vfreex:
@Ximinhan There is nothing we can do if the result count exceeds 1000. It has to be fixed in the server.

vfreex (approved):
/lgtm

PR #450, opened 2022-10-13
[ART-4554] use errata_tool add/remove jira issues

Ximinhan:
REF: https://github.com/openshift/elliott/pull/434

### DIFF
--- a/elliottlib/errata.py
+++ b/elliottlib/errata.py
@@ -635,26 +635,8 @@ def add_jira_bugs_with_retry(advisory_id: int, bugids: List[str], noop: bool = F
         if noop:
             logger.info('Dry run: Would have attached bugs')
             continue
-        results = add_multi_jira_issues(advisory_id, chunk_of_bugs)
-        for i, result in enumerate(results):
-            if result.status_code != 201:
-                rt = add_jira_issue(advisory_id, chunk_of_bugs[i])
-                if rt.status_code != 201:
-                    errata_error = rt.json()
-                    if 'errors' not in errata_error or len(errata_error) != 1:
-                        raise exceptions.ElliottFatalError(f"attach jira bug {chunk_of_bugs[i]} failed with "
-                                                           f"status={rt.status_code} "
-                                                           f"errmsg={rt.json()}")
-                    if 'idsfixed' not in errata_error['errors'] or len(errata_error['errors']) != 1:
-                        raise exceptions.ElliottFatalError(f"attach jira bug {chunk_of_bugs[i]} failed with "
-                                                           f"status={rt.status_code} "
-                                                           f"errmsg={rt.json()}")
-                    for err in errata_error['errors']['idsfixed']:
-                        if 'The issue is filed already in' not in ' '.join(err):
-                            raise exceptions.ElliottFatalError(f"attach jira bug {chunk_of_bugs[i]} failed with "
-                                                               f"status={rt.status_code} "
-                                                               f"errmsg={rt.json()}")
-        logger.info("All not attached jira bugs attached")
+        advisory.addJiraIssues(chunk_of_bugs)
+        advisory.commit()
 
 
 def get_rpmdiff_runs(advisory_id, status=None, session=None):
--- a/elliottlib/errata.py
+++ b/elliottlib/errata.py
@@ -641,7 +641,7 @@ def add_jira_bugs_with_retry(advisory_id: int, bugids: List[str], noop: bool = F
         except ErrataException as e:
             attached_bugs = re.findall("Issue (.*) The issue is filed already in", str(e))
             if attached_bugs:
-                chunk_of_bugs = chunk_of_bugs - [b.upper() for b in attached_bugs]
+                chunk_of_bugs = [b for b in chunk_of_bugs if b not in [b.upper() for b in attached_bugs]]
                 advisory.addJiraIssues(chunk_of_bugs)
                 advisory.commit()
             else:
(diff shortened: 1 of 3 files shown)


vfreex (approved):

elliottlib/errata.py
```diff
@@ -635,26 +635,8 @@ def add_jira_bugs_with_retry(advisory_id: int, bugids: List[str], noop: bool = F
...
-                        raise exceptions.ElliottFatalError(f"attach jira bug {chunk_of_bugs[i]} failed with "
-                                                           f"status={rt.status_code} "
-                                                           f"errmsg={rt.json()}")
-                    for err in errata_error['errors']['idsfixed']:
-                        if 'The issue is filed already in' not in ' '.join(err):
-                            raise exceptions.ElliottFatalError(f"attach jira bug {chunk_of_bugs[i]} failed with "
-                                                               f"status={rt.status_code} "
-                                                               f"errmsg={rt.json()}")
-        logger.info("All not attached jira bugs attached")
+        advisory.addJiraIssues(chunk_of_bugs)
```
thegreyd, line 638: How do "jira already attached" errors get handled here?

Ximinhan: already attached bugs should filtered by #451

Ximinhan:
@thegreyd could you take a look again?

elliottlib/errata.py
```diff
@@ -635,26 +635,17 @@ def add_jira_bugs_with_retry(advisory_id: int, bugids: List[str], noop: bool = F
...
+            advisory.addJiraIssues(chunk_of_bugs)
+            advisory.commit()
+        except ErrataException as e:
+            attached_bugs = re.findall("Issue (.*) The issue is filed already in", str(e))
+            if attached_bugs:
+                chunk_of_bugs = chunk_of_bugs - [b.upper() for b in attached_bugs]
+                advisory.addJiraIssues(chunk_of_bugs)
+                advisory.commit()
+            else:
+                raise exceptions.ElliottFatalError(getattr(e, 'message', repr(e)))
```
vfreex, line 648: I would like to have the original exception raised instead of a new ElliottFatalError.

vfreex (changes requested):

elliottlib/errata.py
Ximinhan: updated

thegreyd:
@Ximinhan I have a few suggestions about the PR, let's get together and discuss, sending a meeting invite

elliottlib/errata.py
```diff
@@ -635,26 +635,17 @@ def add_jira_bugs_with_retry(advisory_id: int, bugids: List[str], noop: bool = F
...
-                                                               f"status={rt.status_code} "
-                                                               f"errmsg={rt.json()}")
-        logger.info("All not attached jira bugs attached")
+        try:
+            advisory.addJiraIssues(chunk_of_bugs)
+            advisory.commit()
+        except ErrataException as e:
+            attached_bugs = re.findall("Issue (.*) The issue is filed already in", str(e))
+            if attached_bugs:
+                chunk_of_bugs = chunk_of_bugs - [b.upper() for b in attached_bugs]
```
vfreex, line 644: I don't think `-` works for list here.

vfreex (changes requested):

elliottlib/errata.py
```diff
@@ -635,26 +635,17 @@ def add_jira_bugs_with_retry(advisory_id: int, bugids: List[str], noop: bool = F
...
-                        if 'The issue is filed already in' not in ' '.join(err):
-                            raise exceptions.ElliottFatalError(f"attach jira bug {chunk_of_bugs[i]} failed with "
-                                                               f"status={rt.status_code} "
-                                                               f"errmsg={rt.json()}")
-        logger.info("All not attached jira bugs attached")
+        try:
+            advisory.addJiraIssues(chunk_of_bugs)
+            advisory.commit()
+        except ErrataException as e:
+            attached_bugs = re.findall("Issue (.*) The issue is filed already in", str(e))
```
vfreex, line 642: I doubt this regex will actually work. `(.*)` is too broad. In case of `re.findall('a (.*) b', 'a 1 b a 2 b')`, it will return `['1 b a 2']` instead of `[1, 2]`.

I don't know what the exception message looks like in case of attached JIRAs. @Ximinhan Did you actually test your PR for a real advisory and JIRAs?

vfreex (changes requested):

elliottlib/errata.py
```diff
@@ -635,26 +635,17 @@ def add_jira_bugs_with_retry(advisory_id: int, bugids: List[str], noop: bool = F
...
+            if attached_bugs:
+                chunk_of_bugs = [b for b in chunk_of_bugs if b not in [b.upper() for b in attached_bugs]]
```
vfreex, line 644: I am not in favor of how the "jira already attached to another advisory" error is handled here.
This is a concept "Garbage in, garbage out". We shouldn't pass already attached jiras to this API at the first place. They should be filtered out before reaching here rather.

vfreex (changes requested):

elliottlib/errata.py
vfreex: Therefore, my opinion is completely removing the logic of detecting "jira already attached to another advisory" error here, then add a filter to prevent already attached jiras from being swept.

elliottlib/errata.py
Ximinhan: I tested, and each bug has a line of error message, the error will looks like `a 1 b\na 2 b`

elliottlib/errata.py
Ximinhan: It was handled in https://github.com/openshift/elliott/pull/451, but seems there will have a performance issue

vfreex: ok, then I think i can merge this PR to test it in the next week. If it doesn't work well I will prefer to revert this and revisit https://github.com/openshift/elliott/pull/451

vfreex (approved):

PR #451, opened 2022-10-13
[ART-4554] use errata_tool filter bug status

Ximinhan:
ref: https://github.com/openshift/elliott/pull/434

### DIFF
--- a/elliottlib/bzutil.py
+++ b/elliottlib/bzutil.py
@@ -22,6 +22,8 @@ from elliottlib import constants, exceptions, exectools, logutil, errata, util
 from elliottlib.cli import cli_opts
 from elliottlib.metadata import Metadata
 from elliottlib.util import isolate_timestamp_in_release
+from errata_tool.jira_issue import JiraIssue as ErrataJira
+from errata_tool.bug import Bug as ErrataBug
 
 logger = logutil.getLogger(__name__)
 
@@ -427,6 +429,10 @@ class BugTracker:
     def get_flaw_bugs(self, bug_ids: List, strict: bool = True, verbose: bool = False):
         raise NotImplementedError
 
+    @staticmethod
+    def filter_attached_bugs(bug_ids: List):
+        raise NotImplementedError
+
 
 class JIRABugTracker(BugTracker):
     @staticmethod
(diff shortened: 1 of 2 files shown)


Ximinhan:
UT failure relies on errata_tool update in https://github.com/openshift/elliott/pull/450

elliottlib/bzutil.py
```diff
@@ -427,6 +429,10 @@ def get_tracker_bugs(self, bug_ids: List, strict: bool = False, verbose: bool =
     def get_flaw_bugs(self, bug_ids: List, strict: bool = True, verbose: bool = False):
         raise NotImplementedError
 
+    @staticmethod
+    def filter_attached_bugs(self, bug_ids: List):
```
thegreyd, line 433: staticmethod won't have self in the arg field

elliottlib/errata.py
```diff
@@ -578,6 +578,8 @@ def add_bugzilla_bugs_with_retry(advisory_id: int, bugids: List, noop: bool = Fa
 
     existing_bugs = bzutil.BugzillaBugTracker.advisory_bug_ids(advisory)
     new_bugs = set(bugids) - set(existing_bugs)
+    # check 150 bugs will take around 5 mins, skip check if there are too many bugs
```
vfreex, line 581: I think if you skip this check, it will be likely to see an error like "Bug X is already attached to another advisory". I would prefer correctness than performance.

vfreex (changes requested):

thegreyd:
I'm not convinced this is an effective solution to handling filtering. Right now it takes atmost 2 requests for us to get all bugs attached - pretty simple - with this approach fetching advisory for each bug - and then optionally relying on secondary logic if number of bugs are really high, is not clean. So @Ximinhan let me know what you think. I personally would recommend to close the PR, and we can revisit this in the future if anything changes

Ximinhan:
This is because when we use the addJiraIssues method, if it fails, then all bugs are not added successfully, and the error information is a summary of all exception strings instead of a json dictionary like before
OK, I will update the exception handling in #450

Ximinhan:
This is because when we use the addJiraIssues method, if it fails, then all bugs are not added successfully, and the error information is a summary of all exception strings instead of a json dictionary like before
OK, I will update the exception handling in #450

Ximinhan:
keep the bug filtering in exception https://github.com/openshift/elliott/pull/450/commits/197b1f2bf4b7d2a114aed467fd6c3833326d5bc1

PR #454, opened 2022-10-21
optimize is_cve_in_summary function

Ximinhan:
simplify return function

### DIFF
--- a/elliottlib/bzutil.py
+++ b/elliottlib/bzutil.py
@@ -61,7 +61,7 @@ class Bug:
         return set(constants.TRACKER_BUG_KEYWORDS).issubset(set(self.keywords))
 
     def is_cve_in_summary(self):
-        return True if re.search(r'^CVE-\d+-\d+', self.summary) else False
+        return bool(re.search(r'^CVE-\d+-\d+', self.summary))
 
     def is_flaw_bug(self):
         return self.product == "Security Response" and self.component == "vulnerability"


elliottlib/bzutil.py
```diff
@@ -61,7 +61,7 @@ def is_tracker_bug(self):
         return set(constants.TRACKER_BUG_KEYWORDS).issubset(set(self.keywords))
 
     def is_cve_in_summary(self):
-        return True if re.search(r'^CVE-\d+-\d+', self.summary) else False
+        return bool(re.search(r'^CVE-\d+-\d+', self.summary))
```
vfreex, line 64: `^CVE-\d+-\d+` will only match strings that start with `CVE-`. I guess it could be better to remove `^` to let it match whatever `CVE-*-*` substring in arbitrary location.

joepvd: hmm not sure if that is required. If prodsec automation had something to do with creating the bug, the first word should start with `CVE`. Do we optimize for false negative or false positives?
I'd be erring on the side of allowing devs to use the word `CVE` somewhere in the summary (and have prodsec fix it up if the tooling did not create the tracker correctly).

joepvd (approved):
/lgtm

elliottlib/bzutil.py
vfreex: This PR is not perfect but it will improve things. So I think I am ok to merge it.

vfreex (approved):

PR #464, opened 2022-11-17
ART-4398 Associate CVE mapping for golang-builder CVEs

thegreyd:
https://issues.redhat.com/browse/ART-4398

## Summary
Right now we are not associating builds with golang CVEs.
With this PR, we will start associating all golang build images
with golang CVEs **except etcd image**. This is because etcd
is built with a different go version (which is behind what our images
use), and in most cases CVEs would be associated with our main go
builder version. So for now a warning is displayed if etcd build is 
detected saying "manually associate etcd build with CVE". 

Handling the etcd case was proving to be quite confusing, so I've
ommitted that to be handled as a followup. This PR gets us much 
closer from where we are today.

## Test
- Test with any advisory that has golang CVE tracker, --noop will still log 
message about golang association.
- Tested it with our 4.12 GA image advisory
- Also with https://errata.devel.redhat.com/advisory/105365/cves
- `elliott --assembly=4.11.16 --group=openshift-4.11 attach-cve-flaws --use-default-advisory image --noop`

### DIFF
--- a/elliottlib/cli/attach_cve_flaws_cli.py
+++ b/elliottlib/cli/attach_cve_flaws_cli.py
@@ -148,10 +148,11 @@ def _update_advisory(runtime, advisory, flaw_bugs, bug_tracker, noop):
 
     flaw_ids = [flaw_bug.id for flaw_bug in flaw_bugs]
     runtime.logger.info(f'Attaching {len(flaw_ids)} flaw bugs')
-    bug_tracker.attach_bugs(advisory_id, flaw_ids, noop)
+    bug_tracker.attach_bugs(flaw_ids, advisory_obj=advisory, noop=noop)
 
 
-async def associate_builds_with_cves(errata_api: AsyncErrataAPI, advisory: Erratum, flaw_bugs: Iterable[Bug], attached_tracker_bugs: List[Bug], tracker_flaws: Dict[int, Iterable], dry_run: bool):
+async def associate_builds_with_cves(errata_api: AsyncErrataAPI, advisory: Erratum, flaw_bugs: Iterable[Bug],
+                                     attached_tracker_bugs: List[Bug], tracker_flaws: Dict[int, Iterable], dry_run: bool):
     attached_builds = [b for pv in advisory.errata_builds.values() for b in pv]
     cve_components_mapping: Dict[str, Set[str]] = {}
     for tracker in attached_tracker_bugs:
@@ -167,7 +168,8 @@ async def associate_builds_with_cves(errata_api: AsyncErrataAPI, advisory: Errat
             cve = flaw_id_bugs[flaw_id].alias[0]
             cve_components_mapping.setdefault(cve, set()).add(component_name)
 
-    await AsyncErrataUtils.associate_builds_with_cves(errata_api, advisory.errata_id, attached_builds, cve_components_mapping, dry_run=dry_run)
+    await AsyncErrataUtils.associate_builds_with_cves(errata_api, advisory.errata_id, attached_builds,
+                                                      cve_components_mapping, dry_run=dry_run)
 
 
 def get_updated_advisory_rhsa(logger, cve_boilerplate: dict, advisory: Erratum, flaw_bugs):
--- a/elliottlib/cli/attach_cve_flaws_cli.py
+++ b/elliottlib/cli/attach_cve_flaws_cli.py
@@ -148,11 +148,10 @@ def _update_advisory(runtime, advisory, flaw_bugs, bug_tracker, noop):
 
     flaw_ids = [flaw_bug.id for flaw_bug in flaw_bugs]
     runtime.logger.info(f'Attaching {len(flaw_ids)} flaw bugs')
-    bug_tracker.attach_bugs(flaw_ids, advisory_obj=advisory, noop=noop)
+    bug_tracker.attach_bugs(advisory_id, flaw_ids, noop)
 
 
-async def associate_builds_with_cves(errata_api: AsyncErrataAPI, advisory: Erratum, flaw_bugs: Iterable[Bug],
-                                     attached_tracker_bugs: List[Bug], tracker_flaws: Dict[int, Iterable], dry_run: bool):
+async def associate_builds_with_cves(errata_api: AsyncErrataAPI, advisory: Erratum, flaw_bugs: Iterable[Bug], attached_tracker_bugs: List[Bug], tracker_flaws: Dict[int, Iterable], dry_run: bool):
     attached_builds = [b for pv in advisory.errata_builds.values() for b in pv]
     cve_components_mapping: Dict[str, Set[str]] = {}
     for tracker in attached_tracker_bugs:
@@ -168,8 +167,7 @@ async def associate_builds_with_cves(errata_api: AsyncErrataAPI, advisory: Errat
             cve = flaw_id_bugs[flaw_id].alias[0]
             cve_components_mapping.setdefault(cve, set()).add(component_name)
 
-    await AsyncErrataUtils.associate_builds_with_cves(errata_api, advisory.errata_id, attached_builds,
-                                                      cve_components_mapping, dry_run=dry_run)
+    await AsyncErrataUtils.associate_builds_with_cves(errata_api, advisory.errata_id, attached_builds, cve_components_mapping, dry_run=dry_run)
 
 
 def get_updated_advisory_rhsa(logger, cve_boilerplate: dict, advisory: Erratum, flaw_bugs):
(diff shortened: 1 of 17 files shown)


elliottlib/util.py
```diff
@@ -544,21 +544,26 @@ def isolate_timestamp_in_release(release: str) -> Optional[str]:
...
         go_version = 'N/A'
         nvr = (build['name'], build['version'], build['release'])
         name = nvr[0]
         if 'golang-builder' in name or 'go-toolset' in name:
             go_version = golang_builder_version(nvr, logger)
-            go_container_nvrs[name] = {
-                'nvr': nvr,
-                'go': go_version
-            }
+            if go_version not in go_nvr_map:
```
Ximinhan, line 564: go_nvr_map.keys()

elliottlib/util.py
```diff
@@ -544,21 +544,26 @@ def isolate_timestamp_in_release(release: str) -> Optional[str]:
     return None
 
 
-def get_golang_container_nvrs(nvrs, logger):
+def get_golang_container_nvrs(nvrs: List[Tuple[str, str, str]], logger) -> Dict[str, Dict[str, str]]:
+    """
+    :param nvrs: a list of tuples containing (name, version, release) in order
+    :param logger: logger
+
+    :return: a dict mapping nvr 'name' to a dict containing nvr and go-builder as values
```
Ximinhan, line 552: :return: a dict mapping go_version to a list contains nvrs

elliottlib/cli/attach_cve_flaws_cli.py
```diff
@@ -149,9 +149,10 @@ def _update_advisory(runtime, advisory, flaw_bugs, bug_tracker, noop):
     bug_tracker.attach_bugs(advisory_id, flaw_ids, noop)
 
 
-async def associate_builds_with_cves(errata_api: AsyncErrataAPI, advisory: Erratum, flaw_bugs: Iterable[Bug], attached_tracker_bugs: List[Bug], tracker_flaws: Dict[int, Iterable], dry_run: bool):
+async def associate_builds_with_cves(errata_api: AsyncErrataAPI, advisory: Erratum, flaw_bugs: Iterable[Bug],
+                                     attached_tracker_bugs: List[Bug], tracker_flaws: Dict[int, Iterable], dry_run: bool):
     attached_builds = [b for pv in advisory.errata_builds.values() for b in pv]
-    cve_components_mapping: Dict[str, Set[str]] = {}
+    cve_components_mapping: Dict[str, Dict] = {}
```
thegreyd, line 155: We now need the flaw bug object to get the information to associate the right builds with each flaw bug.

elliottlib/cli/find_bugs_sweep_cli.py
```diff
@@ -279,7 +279,9 @@ def categorize_bugs_by_type(bugs: List[Bug], advisory_id_map: Dict[str, int], ma
         attached_builds = errata.get_advisory_nvrs(advisory_id_map[kind])
         packages = list(attached_builds.keys())
         if kind == 'image':
-            packages.extend(constants.SPECIAL_CVE_COMPONENTS)
+            # golang builder is a special tracker component
+            # which applies to all our golang images
+            packages.append(constants.GOLANG_BUILDER_CVE_COMPONENT)
```
thegreyd, line 284: Keeping this simple - don't think we're going to have any special cve component except golang builder

elliottlib/cli/get_golang_versions_cli.py
```diff
@@ -72,19 +72,19 @@ def get_nvrs_golang(nvrs, logger):
         else:
             rpm_nvrs.append(nvr_tuple)
 
-    nvrs = {}
     if rpm_nvrs:
-        nvrs.update(util.get_golang_rpm_nvrs(rpm_nvrs, logger))
```
thegreyd, line 77: This was simpler when we were returning nvrs with their go version, now that the method
signatures have changed to returning go_nvr_map - to support this branch - printing image and rpm nvrs 
has become more tedious - instead just print them separate.

elliottlib/errata_async.py
```diff
@@ -1,16 +1,17 @@
...
 import gssapi
+import re
+import semver
 
 from elliottlib.rpm_utils import parse_nvr
-from elliottlib import constants
+from elliottlib import constants, util, logutil
 
-_LOGGER = logging.getLogger(__name__)
+_LOGGER = logutil.getLogger(__name__)
```
thegreyd, line 14: This would let methods here print to our runtime's logger - instead of being lost in the void

elliottlib/cli/verify_attached_bugs_cli.py
```diff
@@ -224,39 +224,44 @@ async def _verify_attached_flaws_for(self, advisory_id: int, attached_trackers:
...
                 cve = flaw_id_bugs[flaw_id].alias[0]
                 cve_components_mapping.setdefault(cve, set()).add(component_name)
-        current_cve_package_exclusions = await AsyncErrataUtils.get_advisory_cve_package_exclusions(self.errata_api, advisory_id)
+
         attached_builds = await self.errata_api.get_builds_flattened(advisory_id)
-        expected_cve_packages_exclusions = AsyncErrataUtils.compute_cve_package_exclusions(attached_builds, cve_components_mapping)
-        extra_cve_package_exclusions, missing_cve_package_exclusions = AsyncErrataUtils.diff_cve_package_exclusions(current_cve_package_exclusions, expected_cve_packages_exclusions)
-        for cve, cve_package_exclusions in extra_cve_package_exclusions.items():
+        try:
+            extra_exclusions, missing_exclusions = await AsyncErrataUtils.validate_cves_and_get_exclusions_diff(
```
thegreyd, line 244: Use existing func instead of repeating code

elliottlib/errata_async.py
```diff
@@ -91,51 +92,114 @@ async def delete_cve_package_exclusion(self, exclusion_id: int):
 
 class AsyncErrataUtils:
     @classmethod
-    async def get_advisory_cve_package_exclusions(cls, api: AsyncErrataAPI, advisory_id: int):
+    async def get_advisory_cve_exclusions(cls, api: AsyncErrataAPI, advisory_id: int):
```
thegreyd, line 95: Var names here are huge and the extra words are not adding to their meaning/intent. Keep them succint

elliottlib/errata_async.py
```diff
@@ -91,51 +92,114 @@ async def delete_cve_package_exclusion(self, exclusion_id: int):
...
+        cve_exclusions = {
+            cve_name: {pkg: 0 for pkg in attached_brew_components - components}
+            for cve_name, components in expected_cve_components.items()
         }
-        return cve_packages_exclusions
+        return cve_exclusions
 
     @classmethod
-    def diff_cve_package_exclusions(cls, current_cve_package_exclusions: Dict[str, Dict[str, int]], expected_cve_packages_exclusions: Dict[str, Dict[str, int]]):
+    def populate_golang_cve_components(cls, golang_cve_names, expected_cve_components, attached_builds):
```
thegreyd, line 144: New logic alert

elliottlib/errata_async.py
```diff
@@ -91,51 +92,114 @@ async def delete_cve_package_exclusion(self, exclusion_id: int):
...
 
     @classmethod
-    async def associate_builds_with_cves(cls, api: AsyncErrataAPI, advisory_id: int, attached_builds: List[str], cve_components_mapping: Dict[str, Set[str]], dry_run=False):
-        """ Request Errata to associate CVEs to attached Brew builds
-        :param api: Errata API
-        :param advisory_id: advisory id
-        :param attached_builds: list of attached Brew build NVRs
-        :param cve_components_mapping: a dict mapping each CVE to a list of brew components
-        """
+    async def validate_cves_and_get_exclusions_diff(cls, api: AsyncErrataAPI, advisory_id: int, attached_builds: List[
```
thegreyd, line 201: Extract out the validation piece so we can reuse it in verify_attached_bugs_cli.py

elliottlib/util.py
thegreyd: Fixed!

elliottlib/util.py
thegreyd: The `in` comparison for dict works as intended (unless I'm missing something)

tests/test_attach_cve_flaws_cli.py
```diff
@@ -87,11 +88,11 @@ async def test_associate_builds_with_cves_bz(self, fake_urils_associate_builds_w
             5: [102],
         }
         attached_tracker_bugs = [
-            BugzillaBug(Mock(id=1, keywords=["Security", "SecurityTracking"], whiteboard="component: a")),
```
thegreyd, line 90: Use constants instead of harcoded values

PR #465, opened 2022-11-22
Verify bugs are not attached to multiple advisories

thegreyd:
We can have a case where a non flaw bug can be attached to multiple advisories

```
curl -sSL --negotiate --user : https://errata.devel.redhat.com/jira_issues/OCPBUGS-35
71/advisories.json | jq -r '.[].advisory_name'
```

Verify that's not the case. Placeholder bugs are excluded so they are not included in this check,
so you won't see the above bug as a complain, but this should catch any other bugs.

Test
- elliott -g openshift-4.11 verify-attached-bugs 105365 --no-verify-blocking-bugs

### DIFF
--- a/elliottlib/bzutil.py
+++ b/elliottlib/bzutil.py
@@ -11,6 +11,8 @@ from datetime import datetime, timezone
 from time import sleep
 from typing import Dict, Iterable, List, Optional
 from jira import JIRA, Issue
+from errata_tool.jira_issue import JiraIssue as ErrataJira
+from errata_tool.bug import Bug as ErrataBug
 
 import bugzilla
 import click
@@ -51,6 +53,10 @@ class Bug:
 
     @property
     def corresponding_flaw_bug_ids(self):
+        raise NotImplementedError
+
+    @property
+    def all_advisory_ids(self):
         raise NotImplementedError
 
     @property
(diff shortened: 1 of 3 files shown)


elliottlib/bzutil.py
```diff
@@ -492,15 +506,15 @@ def get_bugs(self, bugids: List[str], permissive=False, verbose=False, **kwargs)
         if invalid_bugs:
             logger.warn(f"Cannot fetch bugs from a different project (current project: {self._project}):"
                         f" {invalid_bugs}")
-        bugids = [b for b in bugids if self.looks_like_a_jira_project_bug(b)]
+        bugids = {b for b in bugids if self.looks_like_a_jira_project_bug(b)}
```
thegreyd, line 509: This was a bug which was tripping us up on line 518 below

elliottlib/cli/verify_attached_bugs_cli.py
```diff
@@ -60,7 +59,13 @@ async def verify_attached_bugs(runtime: Runtime, verify_bug_status: bool, adviso
         non_flaw_bugs = [b for b in bugs if b.is_ocp_bug()]
 
         validator.validate(non_flaw_bugs, verify_bug_status, no_verify_blocking_bugs)
-        validator.verify_bugs_advisory_type(non_flaw_bugs, advisory_id_map, advisory_bug_map)
+
+        # skip advisory type check if advisories are
+        # manually passed in and we don't know their type
+        if '?' not in advisory_id_map.keys():
```
thegreyd, line 65: do this earlier since `categorize_bugs_by_type` used will trip up if it doesn't find advisory type

Ximinhan:
do we need to add a limit to the length of bug list?

thegreyd:
Don't think we need a limit, I'm testing it on the GA advisories and it's working as expected.

PR #467, opened 2022-11-24
find-bugs:sweep: don't exit with 0 on error

vfreex:
Will try to avoid to directly use exit code. Let's raise an Exception to the top level instead.


thiagoalessio (approved):
/lgtm

PR #471, opened 2022-11-25
Migrate from requests_kerberos to requests-gssapi

vfreex:
`requests-gssapi` is used by koji and errata_tool packages internally, and it works on macOS.

### DIFF
--- a/elliottlib/brew.py
+++ b/elliottlib/brew.py
@@ -15,7 +15,7 @@ from typing import Dict, Iterable, List, Optional, Tuple, BinaryIO
 # 3rd party
 import koji
 import requests
-from requests_kerberos import HTTPKerberosAuth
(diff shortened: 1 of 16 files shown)


vfreex:
If you see unit test fails, it is not related to this PR. PR https://github.com/openshift/elliott/pull/470 should cover the unit test fix.

vfreex:
errata-tool lib no longer raises `spnego.exceptions.GSSError` on authentication failure. We will get a generic `errata_tool.ErrataException` instead.

Updated this PR to remove GSSError captures.

If errata_tool invents a new type dedicated to auth error, we can capture it in our main function to provide more user-friendly error message:
https://github.com/openshift/elliott/blob/94625bf6aa272d1eaa8764994a97da1add478968/elliottlib/cli/__main__.py#L490

PR #474, opened 2022-12-06
fix-drop-advisory

Ximinhan:
move advisory to NEW_FILES before drop builds
test build https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/ximhan/job/build%252Fdrop_advisories/4/console

### DIFF
--- a/elliottlib/cli/advisory_drop_cli.py
+++ b/elliottlib/cli/advisory_drop_cli.py
@@ -24,13 +24,16 @@ def advisory_drop_cli(advisory):
         click.echo(f'Advisory {advisory} is already dropped')
         return
 
+    # move advisory status to NEW_FILES
+    if adv.errata_state != "NEW_FILES":
+        adv.setState("NEW_FILES")
+        adv.commit()
+
     # Remove bugs and builds before dropping the advisory
     all_bugs = adv.errata_bugs
     all_jiras = adv.jira_issues
     all_builds = list(set(itertools.chain.from_iterable(adv.errata_builds.values())))
     if all_bugs or all_jiras or all_builds:
-        if adv.errata_state != "NEW_FILES":
-            adv.setState("NEW_FILES")
         if all_bugs:
             adv.removeBugs(all_bugs)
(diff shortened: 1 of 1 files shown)


elliottlib/cli/advisory_drop_cli.py
```diff
@@ -24,13 +24,17 @@ def advisory_drop_cli(advisory):
         click.echo(f'Advisory {advisory} is already dropped')
         return
 
+    # move advisory status to NEW_FILES
+    if adv.errata_state != "NEW_FILES":
+        adv.setState("NEW_FILES")
+        adv.commit()
+
     # Remove bugs and builds before dropping the advisory
+    adv = Erratum(errata_id=advisory)
```
thegreyd, line 33: I don't think we need this, commit() will probably do an internal refresh. can you test without this?

elliottlib/cli/advisory_drop_cli.py
```diff
@@ -44,7 +48,8 @@ def advisory_drop_cli(advisory):
     data = 'utf8=%E2%9C%93&reason=Dropping+unused+advisory%21&commit=Dropping+unused+advisory'
     headers = {"Content-Type": "text/plain"}
     r = requests.post(url, auth=HTTPSPNEGOAuth(), data=data, headers=headers)
-    if not r.ok:
+    adv = Erratum(errata_id=advisory)
+    if adv.errata_state != "DROPPED_NO_SHIP":
```
thegreyd, line 52: I don't like refetching here, we should trust the return code. I would like to print the return error output if we get any

Ximinhan: the return code is not reliable, see https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/ximhan/job/build%252Fdrop_advisories/3/console it returns 500 but actually advisory already dropped, we don't need to check the return code as long as the advisory status is dropped

joepvd (approved):
/lgtm

PR #480, opened 2022-12-12
WIP Complain when tracker/flaw bugs are inconsistent

thegreyd:


elliottlib/cli/find_bugs_sweep_cli.py
```diff
@@ -256,16 +256,17 @@ def categorize_bugs_by_type(bugs: List[Bug], advisory_id_map: Dict[str, int], ma
 
     tracker_bugs = [b for b in bugs if b.is_tracker_bug() and b.whiteboard_component]
     logger.info(f"Tracker Bugs found: {len(tracker_bugs)}")
+    missed_trackers = [b.id for b in non_tracker_bugs if b.is_cve_in_summary()]
+    if missed_trackers:
+        logger.error(f"Bugs {missed_trackers} have CVE number in summary but do not have tracker labels. Please "
```
thegreyd, line 261: This will error instead of just printing

elliottlib/bzutil.py
```diff
@@ -865,13 +865,21 @@ def id_convert(id_string):
 
     def get_tracker_bugs(self, bug_ids: List, strict: bool = False, verbose: bool = False):
         fields = ["target_release", "blocks", 'whiteboard', 'keywords']
-        return [b for b in self.get_bugs(bug_ids, permissive=not strict, include_fields=fields, verbose=verbose) if
-                b.is_tracker_bug()]
+        bugs = self.get_bugs(bug_ids, permissive=not strict, include_fields=fields, verbose=verbose)
+        trackers = [b for b in bugs if b.is_tracker_bug()]
+        not_trackers = [b.id for b in bugs if not b.is_tracker_bug()]
+        if not_trackers and strict:
+            raise ValueError(f"Unexpected: {not_trackers} were found to not be qualified tracker bugs")
```
thegreyd, line 872: Be strict and complain

elliottlib/cli/find_bugs_sweep_cli.py
```diff
@@ -256,16 +256,17 @@ def categorize_bugs_by_type(bugs: List[Bug], advisory_id_map: Dict[str, int], ma
...
+                     f"fix this")
+
     if not tracker_bugs:
         return bugs_by_type
 
     for b in tracker_bugs:
         logger.info((b.id, b.whiteboard_component))
 
-    warning_bug = [b.id for b in non_tracker_bugs if b.is_cve_in_summary()]
```
thegreyd, line 265: We were missing this by returning early if there were no tracker bugs found

joepvd:
Nice! I like it to shift-left this error to also hit while preparing the advisories, and not hit this error when promoting.

We have fewer methods to override failures in prepare-release, though, we do not have `promotion_permits`. Do you foresee circumstances where we would have put a `promotion_permit`, and now would be stuck? Or how do you think the concrete handling would look like?

PR #491, opened 2023-01-05
[ART-4864] Add microshift advisory support

vfreex:
microshift will be attached to a seperate advisory with impetus `microshift`. This PR

- sweeps microshift related bugs into microshift advisory instead of the default image advisory.
- Adds `--member-only` flag to `find-builds` so only sweep member rpms (rpms in ocp-build-data) .So that microshift can be attached to an advisory using `elliott --group $group --assembly $assembly --rpms microshift find-builds -k rpms --member-only`.


elliottlib/cli/find_bugs_sweep_cli.py
```diff
@@ -250,9 +251,12 @@ def categorize_bugs_by_type(bugs: List[Bug], advisory_id_map: Dict[str, int], ma
...
 
     # for 4.x, first sort all non_tracker_bugs
-    non_tracker_bugs = [b for b in bugs if not b.is_tracker_bug()]
+    non_tracker_bugs = {b for b in bugs if not b.is_tracker_bug()}
     bugs_by_type["extras"] = extras_bugs(non_tracker_bugs)
-    bugs_by_type["image"] = set(non_tracker_bugs) - bugs_by_type["extras"]
+    remaining = non_tracker_bugs - bugs_by_type["extras"]
+    bugs_by_type["microshift"] = {b for b in remaining if b.component and b.component.startswith('MicroShift')}
+    remaining = remaining - bugs_by_type["microshift"]
+    bugs_by_type["image"] = remaining
```
Ximinhan, line 259: bugs_by_type["image"] = remaining - bugs_by_type["microshift"]

elliottlib/cli/find_builds_cli.py
```diff
@@ -366,38 +370,43 @@ def _fetch_builds_by_kind_rpm(runtime: Runtime, tag_pv_map: Dict[str, str], brew
...
-            for component, dep_build in group_deps.items():
-                if component in component_builds and dep_build["id"] != component_builds[component]["id"]:
-                    LOGGER.warning("Swapping stream nvr %s for group dependency nvr %s...", component_builds[component]["nvr"], dep_build["nvr"])
-            component_builds.update(group_deps)
-        builds.extend(component_builds.values())
+
+    if member_only:  # Sweep only member rpms
+        for tag in tag_pv_map:
+            builds_for_tag = asyncio.get_event_loop().run_until_complete(asyncio.gather(*[exectools.to_thread(progress_func, functools.partial(rpm.get_latest_build, default=None, el_target=tag)) for rpm in runtime.rpm_metas()]))
+            builds.extend(filter(lambda b: b is not None, builds_for_tag))
```
Ximinhan, line 378: `builds.extend(list(filter(None, builds_for_tag)))`

elliottlib/cli/find_bugs_sweep_cli.py
vfreex: I saw no difference between those 2 usages.

elliottlib/cli/find_builds_cli.py
vfreex: Personally I prefer to explicitly check if the return value is None here.
Also there is no need to convert the return value to list.

elliottlib/cli/find_builds_cli.py
```diff
@@ -366,38 +370,43 @@ def _fetch_builds_by_kind_rpm(runtime: Runtime, tag_pv_map: Dict[str, str], brew
...
-            # Group dependencies should take precedence over anything previously determined except those pinned by "is".
-            for component, dep_build in group_deps.items():
-                if component in component_builds and dep_build["id"] != component_builds[component]["id"]:
-                    LOGGER.warning("Swapping stream nvr %s for group dependency nvr %s...", component_builds[component]["nvr"], dep_build["nvr"])
-            component_builds.update(group_deps)
-        builds.extend(component_builds.values())
+
+    if member_only:  # Sweep only member rpms
+        for tag in tag_pv_map:
+            builds_for_tag = asyncio.get_event_loop().run_until_complete(asyncio.gather(*[exectools.to_thread(progress_func, functools.partial(rpm.get_latest_build, default=None, el_target=tag)) for rpm in runtime.rpm_metas()]))
```
thegreyd, line 377: Could we split this up into 2 lines? Would like to have readability

PR #498, opened 2023-01-18
bug-sweep: Exclude already attached bugs

vfreex:


joepvd (approved):
/lgtm

Might want to refactor and/or use `get_advisories_for_bug` from `errata.py` and `errata_get_advisories_for_bug_url` from `constants.py`, but we're in a hurry, so am I good with how it is.

vfreex:
> /lgtm
> 
> Might want to refactor and/or use `get_advisories_for_bug` from `errata.py` and `errata_get_advisories_for_bug_url` from `constants.py`, but we're in a hurry, so am I good with how it is.

I deliberately choose the asyncio based implementation because we need to query Errata for every bug and that is somehow slow.

PR #499, opened 2023-01-24
Complain if target version field is missing

thegreyd:
https://redhat-internal.slack.com/archives/G01HDTRUU3T/p1674183594405019


sosiouxme (approved):
should help

PR #502, opened 2023-02-01
[go] Show latest go builds in our build tags for a release

thegreyd:
We've been getting more and more requests asking about our go build version, 
since the new golang cve (https://redhat-internal.slack.com/archives/CB95J6R4N/p1675275965058229)
So I thought it was time we automated some kind of art-bot response. With this
we can hook this up to an art-bot command and ask people to use it (like https://github.com/openshift-eng/art-bot/pull/70)

````
elliott -g openshift-4.12 go
...
Latest go-toolset in rhaos-4.12-rhel-8-build: go-toolset-1.19.4-1.module+el8.8.0+17628+80ec917c brew_buildid=2312278
Latest openshift-golang-builder-container in rhaos-4.12-rhel-8-candidate: openshift-golang-builder-container-v1.19.4-202301042245.el8.g1832cfe brew_buildid=2322066
```

### DIFF
--- a/elliottlib/cli/get_golang_versions_cli.py
+++ b/elliottlib/cli/get_golang_versions_cli.py
@@ -1,6 +1,7 @@
 import click
+import koji
 
-from elliottlib import errata, logutil, util
+from elliottlib import errata, logutil, util, brew, constants
 from elliottlib.cli.common import (cli, find_default_advisory,
                                    use_default_advisory_option)
 from elliottlib.rpm_utils import parse_nvr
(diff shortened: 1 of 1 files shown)


elliottlib/cli/get_golang_versions_cli.py
```diff
@@ -61,6 +64,20 @@ def get_golang_versions_cli(runtime, advisory_id, default_advisory_type, nvrs, c
         nvrs = [n.strip() for n in nvrs.split(',')]
         return get_nvrs_golang(nvrs)
 
+    toolset_name = 'go-toolset'
+    build_tag = f'{runtime.group_config.branch}-build'
```
joepvd, line 68: That is only the default branch. Maybe look at `brew_tag_product_version_mapping` from `erratatool.yml` (and do some `/s/-candidate$/-build/` instead?

vfreex: Joep made a good point.
Another thing is in rhel 7 the package name is `golang-scl-shim`.

PR #507, opened 2023-02-22
Sweep: Don't error on other bugs when a specific advisory type is requested

thegreyd:
This is to enable (https://issues.redhat.com/browse/ART-6108)

```
elliott --group openshift-4.12 --assembly 4.12.5 \
           find-bugs:sweep \
           --use-default-advisory microshift \
           --brew-event $microshift_event \
           --dry-run
```

Right now sweep would error out since it was comparing all found bugs to our advisories. Don't do that when a specific advisory type is requested

### DIFF
--- a/elliottlib/cli/find_bugs_sweep_cli.py
+++ b/elliottlib/cli/find_bugs_sweep_cli.py
@@ -150,10 +150,6 @@ advisory with the --add option.
         logger.info('No bugs found')
         sys.exit(0)
 
-    if output == 'text':
-        click.echo(f"Found {len(bugs)} bugs")
-        click.echo(", ".join(sorted(str(b.id) for b in bugs)))
-
     if report:
         print_report(bugs, output)
 
@@ -181,7 +177,8 @@ async def get_bugs_sweep(runtime: Runtime, find_bugs_obj, brew_event, bug_tracke
     attached_bugs = await bug_tracker.filter_attached_bugs(bugs)
     if attached_bugs:
         attached_bug_ids = {b.id for b in attached_bugs}
-        logger.warning("The following bugs have been attached to advisories: %s", attached_bug_ids)
+        logger.info("These bugs are already attached to some advisory, excluding them: %s",
+                    attached_bug_ids)
         bugs = [b for b in bugs if b.id not in attached_bug_ids]
 
     included_bug_ids, excluded_bug_ids = get_assembly_bug_ids(runtime, bug_tracker_type=bug_tracker.type)
@@ -211,7 +208,7 @@ async def find_and_attach_bugs(runtime: Runtime, advisory_id, default_advisory_t
     bugs = await get_bugs_sweep(runtime, find_bugs_obj, brew_event, bug_tracker)
 
     advisory_ids = runtime.get_default_advisories()
-    bugs_by_type = categorize_bugs_by_type(bugs, advisory_ids,
+    bugs_by_type = categorize_bugs_by_type(bugs, advisory_ids, default_advisory_type,
                                            major_version=major_version)
     for kind, kind_bugs in bugs_by_type.items():
         logger.info(f'{kind} bugs: {[b.id for b in kind_bugs]}')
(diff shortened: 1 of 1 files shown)


elliottlib/cli/find_bugs_sweep_cli.py
```diff
@@ -143,10 +143,6 @@ def find_bugs_sweep_cli(runtime: Runtime, advisory_id, default_advisory_type, ch
         logger.info('No bugs found')
         sys.exit(0)
 
-    if output == 'text':
```
thegreyd, line 146: I don't know in which case this would be useful, in sweep we're printing out bugs by each category. In --use-default-advisory case it lists all bugs found which is confusing, better to remove it

elliottlib/cli/find_bugs_sweep_cli.py
```diff
@@ -174,7 +170,8 @@ def get_bugs_sweep(runtime: Runtime, find_bugs_obj, brew_event, bug_tracker):
     attached_bugs = bug_tracker.filter_attached_bugs(bugs)
     if attached_bugs:
         attached_bug_ids = {b.id for b in attached_bugs}
-        logger.warning("The following bugs have been attached to advisories: %s", attached_bug_ids)
+        logger.info("These bugs are already attached to some advisory, excluding them: %s",
```
thegreyd, line 173: This could be fairly - make it info instead of warning. Further enhancement could be to complain if said advisories are not OCP advisories

elliottlib/cli/find_bugs_sweep_cli.py
joepvd:
```
-o, --output [text|json|slack]  Applies chosen format to --report output
```
I guess it should be `if report and output == "text":
```

File: elliottlib/cli/find_bugs_sweep_cli.py [hunk 3]
```diff
@@ -314,14 +312,19 @@ def categorize_bugs_by_type(bugs: List[Bug], advisory_id_map: Dict[str, int], ma
...
-                   " since no corresponding brew build was found attached to any advisory. "
-                   "First attach builds to the correct advisory and rerun to attach the bugs, "
-                   "or exclude the bug ids in the assembly definition")
-        click.echo(not_found_with_component)
-        raise ValueError(f'No builds found for CVE (bug, package): {not_found_with_component}. Either attach '
-                         f'builds or exclude the bugs in the assembly definition')
+        # Do not error out if a single advisory is requested to be operated on
+        if not default_advisory_type:
+            red_prefix("Tracker Bugs Warning: ")
+            click.echo("The following (tracker bug, package) were found BUT not attached,"
```
joepvd, line 318: Voting for logging rather than `click.echo`. If only because then we dont need to mock the method in unittests.

joepvd (approved):
Nice!

elliottlib/cli/find_bugs_sweep_cli.py
```diff
@@ -314,14 +312,19 @@ def categorize_bugs_by_type(bugs: List[Bug], advisory_id_map: Dict[str, int], ma
...
-        red_prefix("Tracker Bugs Warning: ")
-        click.echo("The following (tracker bug, package) were found BUT not attached,"
-                   " since no corresponding brew build was found attached to any advisory. "
-                   "First attach builds to the correct advisory and rerun to attach the bugs, "
-                   "or exclude the bug ids in the assembly definition")
-        click.echo(not_found_with_component)
-        raise ValueError(f'No builds found for CVE (bug, package): {not_found_with_component}. Either attach '
-                         f'builds or exclude the bugs in the assembly definition')
+        # Do not error out if a single advisory is requested to be operated on
+        if not default_advisory_type:
```
vfreex, line 316: This doesn't seem to be correct. If `--into-default-advisories` flag is set, `default_advisory_type` will be `None` as well and may trigger an error. I think the logic to handle different flag combinations needs to be refactored:
- When `default_advisory_type` is set.
- When `--into-default-advisories` is set.
- When `advisory_id` is set.
- None of above is set.

vfreex (changes requested):

elliottlib/cli/find_bugs_sweep_cli.py
thegreyd: Ah I didn't account for when advisory_id is set. The logic is getting pretty hairy here, accounting for all the 
options. What do you think about deprecating the advisory_id option? Maybe not here but in another PR - I don't think it gets used much - and we have attach-bugs for other cases

vfreex: Maybe this can be refactored like this:
- In `categorize_bugs_by_type`, remove `default_advisory_type` parameter.
- In the caller function, only pass the advisories we'd like to sweep through `advisory_id_map` parameter. e.g. if  `default_advisory_type` is set, we pass `{default_advisory_type: advisory_id}`.
- If `advisory_id` is set. we don't call `categorize_bugs_by_type` at all so no error will be raised.
What do you think?

PR #508, opened 2023-02-23
Switch to https

joepvd:


LICENSE
```diff
@@ -1,6 +1,6 @@
                                  Apache License
                            Version 2.0, January 2004
-                        http://www.apache.org/licenses/
+                        https://www.apache.org/licenses/
```
vfreex, line 3: There are irrelevant changes. Would prefer to keep them.

vfreex (changes requested):

thegreyd:
^ We should test these changes and make sure no cert failures occur

sosiouxme (approved):
/lgtm

i did a test run to fix up 4.12.15, nothing broke

PR #510, opened 2023-02-23
input value check for elliott go cli

Ximinhan:
as we are tying to integrate `elliott go` cli into art-bot, we need to check the input value and return clear output
```
$ ./elliott go -a 00000000
The input value is not valid.
```
```
$ ./elliott go -a ""
Usage: elliott go [OPTIONS]
Try 'elliott go -h' for help.

Error: Invalid value for '--advisory' / '-a': '' is not a valid integer.
```
```
$ ./elliott go -a 111
there is no builds related to golang
```


Ximinhan:
the requirements-dev.txt check failure is not related to the pr, will merge the pr and left the requirements.txt mismatch to synk

PR #511, opened 2023-02-23
ART-6054 Add blocking errata API calls

thegreyd:
https://issues.redhat.com/browse/ART-6054
This is to enable prepare_release.py to set advisory dependencies (read story for details)

To test, in elliott dir, open python shell
```python3
from elliottlib.errata import get_blocking_errata, set_blocking_errata
set_blocking_errata(110351,110349,"SHIPPED_LIVE")
get_blocking_errata(110351)
```
then unset it at https://errata.devel.redhat.com/errata/edit_depends_on/110351

Keep in mind errata endpoints are weird
- The /erratum/{id} response doesn't mention blocking state
- The /blocking_errata_for/{id} response gives advisory in a weird format `{ "2023:0778": "SHIPPED_LIVE"}`. So I'm content to check just the advisory ids from erratum endpoint and trust that the state would be correct.
- The set blocker errata and set blocker state for advisory endpoints are extremely sensitive - the first one 404s if advisory is already listed (?) and second one 500s if advisory isn't added - not mentioning the fact that they are 2 separate endpoints.

So in prepare_release, we'll call get_blocking_errata for rpm, metadata, check image and extra ids are present - if not add them.

### DIFF
--- a/elliottlib/errata.py
+++ b/elliottlib/errata.py
@@ -764,3 +764,39 @@ def is_security_advisory(advisory):
 def is_advisory_impact_smaller_than(advisory, impact):
     i = [None] + constants.SECURITY_IMPACT
     return i.index(advisory.security_impact) < i.index(impact)
+
+def set_blocking_errata(target_advisory_id, blocking_advisory_id, blocking_state):
+    """Set a blocker advisory (at blocking state) for given target advisory
+
+    :param target_advisory_id: advisory number of the target
+    :param blocking_advisory_id: advisory number of the blocker
+    :param blocking_state: a valid advisory state like "SHIPPED_LIVE"
+    """
+    response = ErrataConnector()._post(f'/api/v1/erratum/{target_advisory_id}/add_blocking_errata', data={"blocking_errata": blocking_advisory_id})
+    if response.status_code != requests.codes.ok:
+        raise IOError(f'Failed to set blocking advisory {blocking_advisory_id} for advisory {target_advisory_id} with error: {response.text} status code: {response.status_code}')
+    data = {"blocking_errata": blocking_advisory_id, "blocking_state": blocking_state}
+    response = ErrataConnector()._post(f'/api/v1/erratum/{target_advisory_id}/set_blocker_state_for_blocking_errata', data=data)
+    if response.status_code != requests.codes.ok:
+        raise IOError(f'Failed to set blocking advisory {blocking_advisory_id} for advisory {target_advisory_id} with error: {response.text} status code: {response.status_code}')
+    return response.json()
+
+def get_blocking_errata(advisory_id) -> dict:
+    """Get a list of blocking advisory ids for a given advisory
+    
+    :param advisory_id: advisory number
+    :return: a list of advisory ids
+    """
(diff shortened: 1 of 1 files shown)


elliottlib/errata.py
```diff
@@ -764,3 +764,43 @@ def is_security_advisory(advisory):
 def is_advisory_impact_smaller_than(advisory, impact):
     i = [None] + constants.SECURITY_IMPACT
     return i.index(advisory.security_impact) < i.index(impact)
+
+
+def set_blocking_errata(target_advisory_id, blocking_advisory_id, blocking_state="SHIPPED_LIVE") -> dict:
```
vfreex, line 769: This is not a strong opinion, but I would like to rename it to `set_blocking_advisory`. We call it `advisory` everywhere in our code instead of confusing `erratum` or `errata`, and strictly speaking `errata` is the plural form of `erratum`. Apparently this API call doesn't accept multiple advisories...

elliottlib/errata.py
```diff
@@ -764,3 +764,43 @@ def is_security_advisory(advisory):
...
+        if not "Advisory already listed" in response.text:
+            logger.warning(f'Failed to set blocking advisory {blocking_advisory_id} for advisory {target_advisory_id} with error: {response.text} status code: {response.status_code}')
+    data = {"blocking_errata": blocking_advisory_id, "blocker_state": blocking_state}
+    response = ErrataConnector()._post(f'/api/v1/erratum/{target_advisory_id}/set_blocker_state_for_blocking_errata', data=data)
+    if response.status_code != requests.codes.created:
+        raise IOError(f'Failed to set blocking advisory {blocking_advisory_id} for advisory {target_advisory_id} with error: {response.text} status code: {response.status_code}')
+    return response.json()
+
+
+def get_blocking_errata(advisory_id) -> list:
```
vfreex, line 788: Same here.

elliottlib/errata.py
```diff
@@ -764,3 +764,43 @@ def is_security_advisory(advisory):
...
+    "errata": {
+        "rhba": {
+            "id": 110351,
+            "blocking_advisories": [100, 200]
+            ...
+    """
+    for k in errata:
+        if errata[k]["id"] == advisory_id:
+            return errata[k]['blocking_advisories']
+    return None
```
vfreex, line 806: Could be changed to `[]` to always return a list.

vfreex (changes requested):

elliottlib/errata.py
thegreyd: If blocking_advisories are not found then we have a problem, would like to fail fast

elliottlib/errata.py
thegreyd: Done

elliottlib/errata.py
thegreyd: Changed

elliottlib/errata.py
vfreex: If it is designed to return None in some cases, it would be better to mention it in the documentation, and change the return value type to `Optional[list]`.

thegreyd: updated

PR #522, opened 2023-03-01
ART-6232 Comment on bugs with no target version set

thegreyd:
https://issues.redhat.com/browse/ART-6232

`elliott -g openshift-4.13 find-bugs:notv --comment '<>'`

[context](https://redhat-internal.slack.com/archives/C027ZPSD133/p1677683063810519)

Sample comment using this https://issues.redhat.com/browse/OCPBUGS-7775?focusedCommentId=21828604&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-21828604

### DIFF
--- a/elliottlib/cli/find_bugs_wo_target_version.py
+++ b/elliottlib/cli/find_bugs_wo_target_version.py
@@ -1,0 +1,34 @@
+import json
+import click
+from elliottlib import (Runtime, logutil)
+from elliottlib.cli import common
+from elliottlib.cli.common import click_coroutine
+
+from elliottlib.cli.find_bugs_sweep_cli import FindBugsSweep
+
+
+logger = logutil.getLogger(__name__)
+
+
+@common.cli.command("find-bugs:notv", short_help="Find qualified bugs (sweep) that do not have a target version set")
+@click.option("--comment", "comment",
+              required=False,
+              help="Add comment to found bugs")
+@click.pass_obj
+@click_coroutine
+async def find_bugs_no_tv(runtime: Runtime, comment):
+    runtime.initialize(mode="both")
+    major_version, _ = runtime.get_major_minor()
+    find_bugs_obj = FindBugsSweep()
+
+    bug_tracker = runtime.bug_trackers('jira')
+    bugs = find_bugs_obj.search(bug_tracker_obj=bug_tracker, verbose=runtime.debug,
+                                with_target_release=False, custom_query=' AND "Target Version" is EMPTY AND '
+                                                                        'component != "Release"')
+    print([b.id for b in bugs])
+    # comment = "This bug has been found to have no 'Target Version' field set.
+    # ART automation strictly relies on it to attach bugs to advisories.
+    # Please set the target version to the release that you expect the fix to go in."
+    if comment:
+        for bug in bugs:
+            bug_tracker.add_comment(bug, comment, private=False)
--- a/elliottlib/cli/find_bugs_wo_target_version.py
+++ b/elliottlib/cli/find_bugs_wo_target_version.py
@@ -41,4 +41,4 @@ async def find_bugs_no_tv(runtime: Runtime, comment, report):
 
     if comment:
         for bug in bugs:
-            bug_tracker.add_comment(bug, comment, private=False)
+            bug_tracker.add_comment(bug.id, comment, private=True)
--- a/elliottlib/cli/find_bugs_wo_target_version.py
+++ b/elliottlib/cli/find_bugs_wo_target_version.py
@@ -27,6 +27,7 @@ async def find_bugs_no_tv(runtime: Runtime, comment):
     bugs = find_bugs_obj.search(bug_tracker_obj=bug_tracker, verbose=runtime.debug,
                                 with_target_release=False, custom_query=' AND "Target Version" is EMPTY AND '
                                                                         'component != "Release"')
+    print(f'Found {len(bugs)} bugs with status={find_bugs_obj.status} and no Target Version set')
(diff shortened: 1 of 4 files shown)


elliottlib/cli/find_bugs_wo_target_version.py
```diff
@@ -0,0 +1,54 @@
+import json
+import click
+from elliottlib import (Runtime, logutil)
+from elliottlib.cli import common
+from elliottlib.cli.find_bugs_sweep_cli import FindBugsSweep, print_report
+
+logger = logutil.getLogger(__name__)
+
+
+@common.cli.command("find-bugs:notv", short_help="Find qualified ART managed jira bugs (sweep) that do not have a "
```
vfreex, line 10: Maybe we can make this command more generic, like `find-bugs:invalid`, to detect multiple problems for issues inside OCPBUGS project. e.g.
- Bugs without valid target version
- Reopened bugs
- CVE trackers without required labels or CVE name in the title

We don't have to implement the other 2 checks for now, but it would be nice if we can implement a schedule job to to find invalid bugs in the future to report all problematic bugs. Thoughts?

elliottlib/cli/find_bugs_wo_target_version.py
```diff
@@ -0,0 +1,54 @@
...
+    click.echo(f'Found {len(bugs)} bugs with status={find_bugs_obj.status} and no Target Version set')
+    if report:
+        print_report(bugs)
+    else:
+        click.echo([b.id for b in bugs])
+
+    if comment:
+        logger.info('Adding comment to bugs...')
+        for bug in bugs:
+            bug_tracker.add_comment(bug.id, comment, private=True)
```
vfreex, line 54: I guess we will need a way to avoid duplicate messages.

elliottlib/cli/find_bugs_wo_target_version.py
thegreyd: Really like the idea yuxiang, this can be a catch all for all those invalid cases, and we can fit it in a job. I'll update the PR - rename the command and add some way to specify the checks

elliottlib/cli/find_bugs_wo_target_version.py
thegreyd: I'd like to skip this part for now - reasons being
- we've requested jira folks to add automation at state change 
- also talking with alex/wking re: valid-label req in github prow bot
- and pinged QEs to check for it (aos-devel message, I'll ping on slack as well)
so missing tv could be something that requires our intervention ~ monthly

vfreex: Yeah, we don't have to do this at this moment.

elliottlib/cli/find_bugs_wo_target_version.py
```diff
@@ -0,0 +1,54 @@
...
+@click.option("--comment", "comment",
+              required=False,
+              help="Add comment to found bugs")
+@click.option("--report",
+              required=False,
+              is_flag=True,
+              help="Output a detailed report of found bugs")
+@click.pass_obj
+@common.click_coroutine
+async def find_bugs_no_tv(runtime: Runtime, comment, report):
```
joepvd, line 21: I know what this PR is about, and still i had to spend three seconds imagining what `television` has to do with bugs. Can reduce unnecessary abbreviations?

joepvd (commented):
No objections to the code per se (apart from a naming nit). Think it might be worth going back to the drawing board and figure out how we want such features to behave.

Basically, the PR does two things:
- query jira for issues matching `$requirement`
- act on this search by posting a comment

I can imagine that our needs will change, and that we:
- want to keep more searches that seem to be useful in the future
- have a more general commenting feature

An alternative approach would be to:
- bring jql to find-bugs. This would ensure the proper component matches "under the hood" and allow for something like `elliott bugs --named-query missing_target_version`, where `missing_target_version` would be stored somewhere as a config, maybe in elliott? or in ocp-build-data:main?
- Have a commenter that receives a bunch of bugs on stdin and comments on them

These two facilities would typically be used in a pipe. Example:

```sh
elliott find-bugs --named-query missing_target_version --output ids |
  elliott comment-bugs --comment-file comment.md --not-if-regex "a string we have possibly used in a previous comment"
```

doozer is able to create open bugs these days for PRs it creates: https://github.com/openshift/doozer/blob/59523ee7caefb8456455f29666a10106e631c38b/doozerlib/cli/images_streams.py#L610. This is strictly something different, but at least a related functionality.

elliottlib/cli/find_bugs_wo_target_version.py
thegreyd: That was my "bad (horrible) naming" day, I'll change it

elliottlib/cli/find_bugs_wo_target_version.py
thegreyd: Coming back to this. I took some time to look at verify-bugs command and there is some overlap with this. It already does some of the trackers without labels / cve name in the title validation.
Filed https://github.com/openshift-eng/elliott/pull/529 to update the help

Target release is really a special case, instead of shoe-horning this and verify-bugs, I really like @joepvd 's idea - https://github.com/openshift-eng/elliott/pull/522#pullrequestreview-1328405875. So we separate out named queries in yml and then keep a generic search  to search and then another to comment. I'll write up a story for that

thegreyd:
Closing in favor of https://issues.redhat.com/browse/ART-6293 and https://issues.redhat.com/browse/ART-6368

PR #523, opened 2023-03-07
add bug's url in message

Ximinhan:


elliottlib/cli/verify_attached_bugs_cli.py
```diff
@@ -386,13 +386,13 @@ def _verify_bug_status(self, bugs):
             status = f"{bug.status}"
             if bug.status in ['CLOSED', 'Closed']:
                 status = f"{bug.status}: {bug.resolution}"
-            self._complain(f"Bug {bug.id} has status {status}")
+            self._complain(f"Bug <{bug.weburl}|{bug.id}> has status {status}")
```
vfreex, line 389: BugzillaBug doesn't seem to have `weburl`.

vfreex (changes requested):

elliottlib/cli/verify_attached_bugs_cli.py
Ximinhan: bugzilla object has weburl https://github.com/python-bugzilla/python-bugzilla/blob/main/bugzilla/bug.py#L42

vfreex: But `BugzillaBug` is not of type `bugzilla.Bug`.

vfreex (approved):

PR #524, opened 2023-03-07
Fix an issue in bug cutoff timestamps estimate

vfreex:
Elliott uses timestamps encoded in image NVRs to estimate the bug cutoff timestamp when sweeping bugs with a Brew event.

When there is a new image added to ocp-build-data, the build might not exist at the moment of the Brew event. Then we will see an error like:

```
OSError: No builds detected for using prefix: 'nmstate-console-plugin-container-v4.13.', extra_pattern: '*', assembly: 'stream', build_state: 'COMPLETE', el_target: 'None'
```

We need to update `bzutil.approximate_cutoff_timestamp` to ignore non-existent builds when estimating cutoff timestamps.


joepvd (approved):
/lgtm

PR #531, opened 2023-03-22
[ART-6354] Setup synk

ashwindasr:
Setup snyk (https://issues.redhat.com/browse/ART-6354)

Steps
- Get the github token by following this [doc](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/authenticating-as-a-github-app-installation) using openshift-art-build-bot credentials on bitwarden. Generate a JWT token first and using that generate a github token
- Add a secret to the repo by using GitHub API endpoint ([docs](https://docs.github.com/en/rest/actions/secrets?apiVersion=2022-11-28#create-or-update-a-repository-secret)) _(but make sure to check if a secret with the same name doesn't exist, else the new value will overwrite the old one!)_
  - You will need the repository public key, which you can get from [here](https://docs.github.com/en/rest/actions/secrets?apiVersion=2022-11-28#get-a-repository-public-key)
  - And also the encrypt the secret value using the instructions given in the [doc](https://docs.github.com/en/rest/actions/secrets?apiVersion=2022-11-28#create-or-update-a-repository-secret).

Test run: https://github.com/ashwindasr/art-bot/actions/runs/4492999119
_(Test run does not inclue `synk code` command as it needs to be enabled in the synk dashboard [settings](https://app.snyk.io/org/openshift-art-build-bot/manage/snyk-code), which has been done for openshift-eng)_

### DIFF
--- a/.github/workflows/snyk_security.yaml
+++ b/.github/workflows/snyk_security.yaml
@@ -1,0 +1,21 @@
+name: Snyk Security
+
+on: [ push, pull_request ]
+
+jobs:
+  snyk:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: Test for open source vulnerabilities and license issues.
+        uses: snyk/actions/python-3.8@master
+        env:
+          SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
+        with:
+          command: test
+      - name: Test for any known security issues using Static Code Analysis.
+        uses: snyk/actions/python-3.8@master
+        env:
+          SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
+        with:
+          command: code test
--- a/.github/workflows/snyk_security.yaml
+++ b/.github/workflows/snyk_security.yaml
@@ -1,4 +1,4 @@
-name: Snyk Security
+name: Snyk Scan
 
 on: [ push, pull_request ]
 


ashwindasr:
Synk test failing because its trying to pull the secret from my fork repo secrets. Once merged, it'll look for the secret defined in the openshift-eng/elliott repo

.github/workflows/snyk_security.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+on: [ push, pull_request ]
+
+jobs:
+  snyk:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: Test for open source vulnerabilities and license issues.
+        uses: snyk/actions/python-3.8@master
```
thegreyd, line 11: I guess we don't care about this action being python-3.8 since this wouldn't be invoking elliott in any way. This can be the latest version of python / what the action supports

thegreyd: Or am I confusing this.. does it include 3.8 relevant things in scan maybe?

ashwindasr: No, I don't think it matters. Tested it on my art-bot fork. Its working.

thegreyd: https://github.com/snyk/actions/blob/806182742461562b67788a64410098c9d9b96adb/python-3.8/action.yml#L22 looks to be the important part that we want - basically communicate to snyk to look for 3.8 relevant vulnerabilities

ashwindasr: Ahh I see. Okay I'll add that change in.

thegreyd: Things would break with python version above 3.8 - I don't know if pip would break. I am good with using the 3.8 action

thegreyd:
Another part of the story is to not fail if for some reason snyk cannot run due to quota issues. For now what we can do is not make this be a required check on PRs - and then work around this issue when we hit it

.github/workflows/snyk_security.yaml
ashwindasr: I did a test run and it didn't break. But since 3.8 is our production env, its better that we keep it that way I guess

PR #533, opened 2023-03-25
Fix devcontainer to use pyproject.toml

vfreex:
Newer versions of pip and setuptools are required to use `pip install -e`


sosiouxme (approved):
thanks, works for me

PR #535, opened 2023-03-27
Ignore when new jira bugs are not found in ET

thegreyd:
We are doing an ET lookup for all found jira bugs - newly created jira bugs
will 404 since they are not sync'd. Handle that error instead of crashing

Tested with: `./elliott --data-path ../ocp-build-data -g openshift-4.14 --assembly stream find-bugs:sweep`
(without it) `2023-03-27 11:02:30,820 ERROR exception with jira bug tracker: 404, message='Not Found', url=URL('https://errata.devel.redhat.com/jira_issues/OCPBUGS-10908/advisories.json')`

### DIFF
--- a/elliottlib/errata_async.py
+++ b/elliottlib/errata_async.py
@@ -2,6 +2,7 @@ import asyncio
 import base64
 from typing import Dict, Iterable, List, Set, Union
 from urllib.parse import quote, urlparse
+from aiohttp import ClientResponseError
 
 import aiohttp
 import gssapi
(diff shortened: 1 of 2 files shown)


elliottlib/errata_async.py
```diff
@@ -91,9 +92,19 @@ async def delete_cve_package_exclusion(self, exclusion_id: int):
...
-        return await self._make_request(aiohttp.hdrs.METH_GET, path)
+        try:
+            result = await self._make_request(aiohttp.hdrs.METH_GET, path)
+        except ClientResponseError as e:
+            # When newly created jira bugs are not sync'd to ET we get a 404, 
+            # assume that they are not attached to any advisory
+            if ignore_not_found and e.status == 404:
+                result = []
+            else:
+                raise(e)
```
locriandev, line 105:
```suggestion
                raise
```

locriandev (commented):
LGTM, just suggested a tiny cosmetic change

PR #543, opened 2023-04-24
Support showing go version for releases/nightlies

thegreyd:
Since we're supporting different go versions in a payload with canonical_builders_from_upstream config option,
it's useful to get builder info for all images in a payload to compare.

I extracted out code from verify-payload to get nvrs from payload, so it could be used here as well.

Test:
- `elliott -g openshift-4.14 go -r registry.ci.openshift.org/ocp/release:4.14.0-0.nightly-2023-04-24-145153`
- `elliott -g openshift-4.12 --assembly 4.12.14 verify-payload quay.io/openshift-release-dev/ocp-release:4.12.14-x86_64 113026`

### DIFF
--- a/elliottlib/cli/__main__.py
+++ b/elliottlib/cli/__main__.py
@@ -12,14 +12,13 @@ web service.
 # stdlib
 from multiprocessing import cpu_count
 from multiprocessing.dummy import Pool as ThreadPool
-import asyncio
 import datetime
 import json
 import sys
 from typing import Dict, List
 
 # ours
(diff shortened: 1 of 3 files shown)


joepvd:
Very nice! Looks mergable to me as soon as the tests succeed.

Two ideas that crossed my mind while reading this:
- Maybe a useful additional switch would be `--not-in-stream` or similar, which would only output the result if the golang builder is not in streams.yml for that version
- Maybe the output would be better in yaml, either behind a switch or by default?

PR #548, opened 2023-05-04
[verify-bugs] Don't error if a bug has whiteboard component defined

thegreyd:
Bugs other than tracker bugs can have whiteboard components defined - currently we are detecting them as invalid trackers but they aren't. (https://bugzilla.redhat.com/show_bug.cgi?id=2092289)


vfreex (approved):
/lgtm

PR #551, opened 2023-05-12
Add `find-unused-rpms` to find unused rpms

vfreex:
This command will list all tagged rpms that are not used in RHCOS or any images.


joepvd (approved):
/lgtm

PR #553, opened 2023-05-25
Move to another API to list advisory builds

vfreex:
API /api/v1/erratum/{id}/builds_list seems to be problematic. With advisory 114782 it is missing one build.

Use /api/v1/erratum/{id}/builds instead.


thiagoalessio (approved):
/lgtm

locriandev:
Is this meant to be permanent? Noticed you haven't bumped elliott yet

vfreex:
@locriandev yes, it is permanent. Errata has some "duplicated" API endpoints and we can use another API endpoint to work around this.

Elliott will be bumped with Doozer on Friday.

PR #556, opened 2023-06-02
errata_async: use per-request GSSAPI token

vfreex:
Currently a GSSAPI token is generated when `login` method is called and that token is reused for all following API requests. This approach has been broken since the recent Errata Tool update.

Moving to per-request token should solve this issue.


thegreyd:
I tested it locally and it works! thanks @vfreex !
/lgtm

PR #561, opened 2023-06-09
find-bugs:qe - Use local logger instead of click.echo

thegreyd:
stdout isn't being forwarded from elliott find-bugs:qe in ocp4 job,
so use logger instead.

slack ref: https://redhat-internal.slack.com/archives/GDBRP5YJH/p1686309961426089?thread_ts=1686302994.401709&cid=GDBRP5YJH


joepvd (approved):
/lgtm

PR #567, opened 2023-06-30
use hack errata_tool

Ximinhan:
advisory schema changed, errata_tool need update


joepvd (approved):

joepvd (changes requested):
Think a pin to a specific commit is better than pointing to the default branch. Like [here](https://github.com/openshift-eng/elliott/pull/383/files)

joepvd (approved):
Thanks!
/lgtm

PR #568, opened 2023-06-30
Add Move builds command

thegreyd:
Sometimes we need the ability to simply move builds from one advisory to another
exactly without needing to worry about reanalyzing brew/tags/ET and an extra build slipping in
, like when preparing optional operator advisory at GA time or otherwise in special cases

`elliott move-builds --from 123 --to 456 -k image --noop`

This was tested with this week's release advisories (moving builds from extras -> image)


sosiouxme (commented):
for the silent ops release, we don't want to move everything from an advisory, just a subset... would be nice to specify that?

it's a shame we don't have a concise way to represent the desired outcome in an assembly (and just re-run prepare-release)

tests would be nice...

sosiouxme (approved):
/lgtm

PR #572, opened 2023-07-07
Retry jira search when searching for kernel bugs

thegreyd:
Jira search has a tendency to flake once in a while and throw a 401, take care of that


joepvd (approved):
/lgtm

PR #573, opened 2023-07-11
[ART-6668] don't sweep microshift bug without check

Ximinhan:
* revert the workaround of https://github.com/openshift-eng/elliott/pull/558

for microshfit bug, don't attach and don't raise error if microshift advisory has no build
this change won't sweep microshift bugs by default so need to find bugs for microshift advisory after build attached, which included in https://issues.redhat.com/browse/ART-6669 with https://github.com/openshift-eng/aos-cd-jobs/pull/3781

### DIFF
--- a/elliottlib/cli/find_bugs_sweep_cli.py
+++ b/elliottlib/cli/find_bugs_sweep_cli.py
@@ -308,30 +308,30 @@ def categorize_bugs_by_type(bugs: List[Bug], advisory_id_map: Dict[str, int], ma
             continue
         attached_builds = errata.get_advisory_nvrs(advisory)
         packages = list(attached_builds.keys())
-        exception_packages = []
         if kind == 'image':
             # golang builder is a special tracker component
             # which applies to all our golang images
-            exception_packages.append(constants.GOLANG_BUILDER_CVE_COMPONENT)
-
-        if kind == 'microshift':
-            # microshift is special since it has a separate advisory, and it's build is attached
-            # after payload is promoted. So do not pre-emptively complain
-            exception_packages.append('microshift')
+            packages.append(constants.GOLANG_BUILDER_CVE_COMPONENT)
 
         for bug in tracker_bugs:
             package_name = bug.whiteboard_component
-            if (package_name in packages) or (package_name in exception_packages):
+            if (package_name in packages):
                 if package_name in packages:
                     logger.info(f"{kind} build found for #{bug.id}, {package_name} ")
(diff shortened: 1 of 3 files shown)


elliottlib/cli/find_bugs_sweep_cli.py
```diff
@@ -308,30 +308,30 @@ def categorize_bugs_by_type(bugs: List[Bug], advisory_id_map: Dict[str, int], ma
...
         if kind == 'image':
             # golang builder is a special tracker component
             # which applies to all our golang images
-            exception_packages.append(constants.GOLANG_BUILDER_CVE_COMPONENT)
-
-        if kind == 'microshift':
-            # microshift is special since it has a separate advisory, and it's build is attached
-            # after payload is promoted. So do not pre-emptively complain
-            exception_packages.append('microshift')
+            packages.append(constants.GOLANG_BUILDER_CVE_COMPONENT)
```
vfreex, line 314: Why is `exception_packages` removed? I think it is better to keep it and make our logs clear about it is an exception.

elliottlib/cli/find_bugs_sweep_cli.py
```diff
@@ -308,30 +308,30 @@ def categorize_bugs_by_type(bugs: List[Bug], advisory_id_map: Dict[str, int], ma
...
-                    logger.info(f"{package_name} bugs included by default")
                 found.add(bug)
                 bugs_by_type[kind].add(bug)
 
     not_found = set(tracker_bugs) - found
-    if not_found:
-        not_found_with_component = [(b.id, b.whiteboard_component) for b in not_found]
+    not_found_with_component = []
+    for b in not_found:
+        if b.whiteboard_component == "microshift" and len(errata.get_advisory_nvrs(advisory_id_map.get("microshift"))) == 0:
```
vfreex, line 327: Duplicate Errata API call for the same advisory.

vfreex (changes requested):

elliottlib/cli/find_bugs_sweep_cli.py
Ximinhan: pr updated, keep golang bug in exception

elliottlib/cli/find_bugs_sweep_cli.py
Ximinhan: pr updated, move this check to bug checking loop

vfreex (approved):

vfreex:
LGTM

PR #575, opened 2023-07-21
remove setuptools-sm for art-tools merger

ashwindasr:
When we do pip install for elliott in merged art-tools repo, setuptools requires it to be a git repo (submodule), which elliott will no longer be.


sosiouxme (approved):
/lgtm

PR #577, opened 2023-07-27
Don't error out if CVE has a non-CVE alias

Ximinhan:
align with https://github.com/openshift-eng/elliott/pull/569


joepvd (approved):
/lgtm

PR #581, opened 2023-08-02
ART-7268 Say in bug that it will not be visible in advisory if so

locriandev:
Ref. https://issues.redhat.com/browse/ART-7268

- rename `runtime.bug_trackers` into `runtime.get_bug_tracker`; the function returns a single object, and it's not a property so making this clear in the name. Also added return type hint: https://github.com/openshift-eng/elliott/commit/3a43d552a7b5d61b77b7ae2056b7109ae92a4f68
- when moving MODIFIED bugs to QE, if the bug visibility is by any means restricted, add a comment stating that the bug won't be visible in the advisory: https://github.com/openshift-eng/elliott/commit/9f8321303caff6d15e59e491bae77f3b890cf03b

### DIFF
--- a/elliottlib/bzutil.py
+++ b/elliottlib/bzutil.py
@@ -227,6 +227,13 @@ class JIRABug(Bug):
     def status(self):
         return self.bug.fields.status.name
 
+    @property
+    def security_level(self):
+        try:
+            return self.bug.fields.security
+        except AttributeError:
+            return None
+
     def is_tracker_bug(self):
         has_keywords = set(constants.TRACKER_BUG_KEYWORDS).issubset(set(self.keywords))
         has_whiteboard_component = bool(self.whiteboard_component)
--- a/elliottlib/cli/find_bugs_qe_cli.py
+++ b/elliottlib/cli/find_bugs_qe_cli.py
@@ -59,11 +59,21 @@ def find_bugs_qe(runtime, find_bugs_obj, noop, bug_tracker):
         f" expected in the next created {major_version}.{minor_version} nightly and release.")
     for bug in bugs:
         updated = bug_tracker.update_bug_status(bug, 'ON_QA', comment=release_comment, noop=noop)
-        if updated and bug.is_tracker_bug():
-            # leave a special comment for QE
-            comment = """Note for QE:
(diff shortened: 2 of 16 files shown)


joepvd (commented):
lgtm, apart from test failure :)

elliottlib/bzutil.py
```diff
@@ -227,6 +227,13 @@ def component(self):
     def status(self):
         return self.bug.fields.status.name
 
+    @property
+    def security_level(self):
+        try:
+            return self.bug.fields.security
```
thegreyd, line 233: Curious what this returns when the bug is public, False / doesn't exist?

elliottlib/cli/find_bugs_qe_cli.py
```diff
@@ -35,7 +35,7 @@ def find_bugs_qe_cli(runtime: Runtime, noop):
     runtime.initialize()
     find_bugs_obj = FindBugsQE()
     exit_code = 0
-    for b in [runtime.bug_trackers('jira'), runtime.bug_trackers('bugzilla')]:
+    for b in [runtime.get_bug_tracker('jira'), runtime.get_bug_tracker('bugzilla')]:
```
thegreyd, line 38: We can remove `bugzilla` from our bug lookup commands (find-bugs:qe, find-bugs:sweep), probably better in a different PR

elliottlib/bzutil.py
locriandev: If the bug is public there won't be a security attribute. Code will catch the exception and return None.

elliottlib/cli/find_bugs_qe_cli.py
locriandev: Sounds good. If you could open a ticket, I can take a look

thegreyd:
I'm not too hung up on that part, since right now we do the same with other comments (post everytime we move to qe). So not a blocker imo

thegreyd (approved):

locriandev:
> I'm not too hung up on that part, since right now we do the same with other comments (post everytime we move to qe). So not a blocker imo

Agree, I thought we could afford repeating ourselves in these cases

PR #583, opened 2023-08-02
Test updates to PRs

joepvd:


locriandev (approved):

thegreyd:
I don't think we need this since 
`By default, a workflow only runs when a pull_request event's activity type is opened, synchronize, or reopened.`
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

PR #584, opened 2023-08-03
Remove bugzilla search from sweep commands

thegreyd:
Normal OCP bugs are no longer allowed to open in Bugzilla.
Bugzilla is only used for tracking flaw bugs for OCP purposes (which are not part of sweep commands)
So remove searching bugzilla for ocp bugs

Test
- `./elliott --group openshift-4.13 --assembly 4.13.8 find-bugs:sweep --use-default-advisory microshift --noop`

### DIFF
--- a/elliottlib/cli/find_bugs_blocker_cli.py
+++ b/elliottlib/cli/find_bugs_blocker_cli.py
@@ -60,15 +60,8 @@ Use --exclude_status to filter out from default status list.
     find_bugs_obj = FindBugsBlocker()
     find_bugs_obj.include_status(include_status)
     find_bugs_obj.exclude_status(exclude_status)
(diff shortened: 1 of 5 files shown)


locriandev (commented):
LGTM, but why removing exception handling?

thegreyd:
The try-catch block was useful when there were multiple bug trackers being run - so an error from one wouldn't fail the whole command - we would print it out and then continue. Now that it's just one bug tracker, any exception should end the command in failure which is fine since nothing else is running in parallel.. I am fine keeping it as well since we put in place these abstractions to support multiple bug trackers at any time

thegreyd:
Ack. Let's park this PR until we have confirmation

